PR #6, opened 2024-06-11
migrate to new validator

karlnaden:
# Summary

updated to use new validator

# Testing Guidance

use the suite instructions to test
1. SMART app using postman
2. Payer server against SMART app tests


rpassas (approved):
Tests behave as expected

PR #7, opened 2024-06-18
Fi 2821 payer spec tests

rpassas:
# Summary
- Uses a shared context for reducing redundant code for each set of tests
- Implements initial wait tests, request and response validation, and cql library check
- does not implement $next-question validation for adaptive flow or extension checks (may need better resources to test against)

# Testing Guidance
run spec tests

### Notes
Can squash commits if necessary.

Commented out a line for response validation - for some reason I had trouble with stubbing validation request for a bundle, but not for parameters


spec/davinci_dtr_test_kit/shared_setup.rb
```diff
@@ -0,0 +1,23 @@
+RSpec.shared_context('when payer server static') do
+  let(:group) { Inferno::Repositories::TestGroups.new.find('payer_server_static_package') }
+  let(:suite_id) { :dtr_payer_server }
+  let(:questionnaire_package_url) { "/custom/#{suite_id}/fhir/Questionnaire/$questionnaire-package" }
+  let(:session_data_repo) { Inferno::Repositories::SessionData.new }
+  let(:test_session) { repo_create(:test_session, test_suite_id: suite_id) }
```
karlnaden, line 6: Lots of stuff in here that is going to be test-specific and so not sharable. Is there a way to pass some of these details in - group and suite id in particular to make this generic across those - or is that fundamentally counter to how rspec works?

If we need to define them statically, then maybe an approach where you use a test file to generate this boilerplate at the top of your unit tests for that specific test...

rpassas: Rsponding now that you are back - shared context creates shared "stuff" that can be modified up front. So, at the top of the spec file, I can pass the context some parameters to dictate what `url` to use, for example. 

I try to organize the tests by flow to try to keep it organized.


Screenshot 2024-07-01 at 10 01 36 PM

PR #12, opened 2024-07-09
rm extra assert

rpassas:
# Summary
- refactoring request validation. The `perform_request_validation_test` method is only really useful when there are multiple requests to validate. This is only the case when taking multiple next-question requests.
- updating the request validation spec tests (adjusting stubs)
- adjusted how messages were presented for readability
- some general clean up of text (e.g. [User Input Validation] removed), some duplicate text deleted

# Testing Guidance

Run server tests against the Smart App tests
Run spec tests

To see updated messages run static or adaptive tests against the server. The main change in presentation is that only next-question requests/responses and the checks on adaptive responses (e.g. 2.07 + 2.08) should indicate the [Resource X] label since the other requests/responses should be singular.

**note:** I do notice that one of the response validation fails with the new validator running the static server tests against the smart app suite. Might be an issue with the response resource - It is not finding the questionnairePackageBundle resource

### DIFF
--- a/lib/davinci_dtr_test_kit/payer_server_groups/payer_server_adaptive_response_bundles_validation_test.rb
+++ b/lib/davinci_dtr_test_kit/payer_server_groups/payer_server_adaptive_response_bundles_validation_test.rb
@@ -33,8 +33,7 @@ module DaVinciDTRTestKit
         end
       end
       if !test_passed && !tests_failed[profile_url].blank?
-        raise assert test_passed,
-                     "Not all returned resources conform to the profile: #{profile_url}"
+        assert test_passed, "Not all returned resources conform to the profile: #{profile_url}"
       end
     end
   end
--- a/lib/davinci_dtr_test_kit/payer_server_groups/payer_server_adaptive_response_bundles_validation_test.rb
+++ b/lib/davinci_dtr_test_kit/payer_server_groups/payer_server_adaptive_response_bundles_validation_test.rb
@@ -20,19 +20,22 @@ module DaVinciDTRTestKit
 
     run do
       skip_if retrieval_method == 'Static', 'Performing only static flow tests - only one flow is required.'
-      test_passed = false
+      test_passed = true
       profile_url = 'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/DTR-QPackageBundle|2.0.1'
       assert !scratch[:adaptive_responses].nil?, 'No resources to validate.'
       scratch[:adaptive_responses].each_with_index do |resource, index|
         fhir_resource = FHIR.from_contents(resource.response[:body])
         fhir_resource.parameter.each do |param|
           resource_is_valid = validate_resource(param.resource, :bundle, profile_url, index)
-          test_passed = true if resource_is_valid
+          test_passed = false unless resource_is_valid
         rescue StandardError
           next
         end
       end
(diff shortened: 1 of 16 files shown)


karlnaden (commented):
why did you remove the unit test?

rpassas:
> why did you remove the unit test?

As it is now, it fails with this change.

lib/davinci_dtr_test_kit/payer_server_groups/payer_server_adaptive_response_bundles_validation_test.rb
```diff
@@ -21,7 +21,7 @@ class PayerAdaptiveFormResponseBundlesTest < Inferno::Test
     run do
       skip_if retrieval_method == 'Static', 'Performing only static flow tests - only one flow is required.'
       test_passed = false
```
karlnaden, line 23: looking at this logic, it seems like it will pass if at least one of the adaptive responses is conformant. Don't we want all of them to be conformant? I think it would look like initializing test_passed to true and then setting it in the loop to `test_passed && resource_is_valid`, right?

lib/davinci_dtr_test_kit/payer_server_groups/payer_server_adaptive_response_search_validation_test.rb
```diff
@@ -34,9 +34,8 @@ class PayerAdaptiveFormResponseSearchTest < Inferno::Test
           next
         end
       end
-      raise tests_failed[profile_url][0] if !test_passed && !tests_failed[profile_url].blank?
+      raise assert test_passed, "No returned resources conform to the profile: http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/dtr-questionnaire-adapt-search" if !test_passed && !tests_failed[profile_url].blank?
```
karlnaden, line 37: 1. Should this have a version (here and above) like you added to others? Seems like yes.
2. The combination or raise and assert looks wrong

karlnaden:
regarding the error, it appears that there is some inconsistent output parameter naming between the [operation](https://hl7.org/fhir/us/davinci-dtr/STU2/OperationDefinition-questionnaire-package.html) where it is called "return" and the [output parameters](https://hl7.org/fhir/us/davinci-dtr/STU2/StructureDefinition-dtr-qpackage-output-parameters.html) where it is called "PackageBundle". We'll try to enter that as a ticket in their upcoming (offcyle I think) ballot, but in the meantime, let's make sure that our suites are consistent.  Since the our payer server simulation seems to be returning the [operation](https://hl7.org/fhir/us/davinci-dtr/STU2/OperationDefinition-questionnaire-package.html) form, our validation logic should assume the same, which means validating against the profiles used on the parameter entries rather than the parent parameters instance - does that make sense?

karlnaden (changes requested):
Generally looks good - a few questions and requested fixes

lib/davinci_dtr_test_kit/payer_server_groups/payer_server_adaptive_response_bundles_validation_test.rb
rpassas: I see what you mean. I think this either an oversight or initially done because the RI would return multiple questionnaires and only one was of interest anyway. But we should not test to an implementation of course. I will flip the logic

karlnaden (approved):

PR #15, opened 2024-09-06
Full ehr tests

karlnaden:
# Summary

- Basic full ehr test suite based off of the smart app suite, but using mostly attestations

# Testing Guidance

- use the postman-based demo instructions in the [suite documentation](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/full-ehr-tests/lib/davinci_dtr_test_kit/docs/dtr_full_ehr_suite_description_v201.md#postman-based-demo) to verify that the tests work as expected
- review documentation


rpassas:
I was going to suggest using Epic's response as the default, but it's a bit silly and it probably makes sense to keep with the static dinner theme for now.

karlnaden:
> I was going to suggest using Epic's response as the default, but it's a bit silly and it probably makes sense to keep with the static dinner theme for now.

Its an open question whether we should allow the tester to provide the questionnaire like we do with other tests, but I don't want to make that change now and I definitely don't want to build the tests around a questionnaire provided by a specific vendor.

lib/davinci_dtr_test_kit/docs/dtr_full_ehr_suite_description_v201.md
```diff
@@ -0,0 +1,127 @@
...
+1. Attest to the remainder of the tests as desired to get a sense for what is involved in testing
+   with an actual EHR implementation. To see what a valid QuestionnaireResponse looks like, see
+   the "Sample QuestionnaireResponse for Dinner (Static) ..." request in postman.
+
+## Limitations
+
+The DTR IG is a complex specification and these tests currently validate conformance to only
+a subset of IG requirements. Future versions of the test suite will test further
+features. A few specific features of interest are listed below.
+
```
rpassas, line 115: Could not that while Full EHR SHALL require support of adaptive questionnaires as well (unlike payer servers), we are only covering static ones for now. It might also be good to put caveats for items we are not testing on the client side. I may have been confused by the IG here and included things that apply to only the SMART app - I am not sure how liberally the authors mean to use 'client system' or 'DTR client'.

Other items that are not covered:
- client SHALL visually flag any elements that require adjustment to meet validation rules (or is this SMART APP specific?)
- DTR applications SHALL support rendering according to the mustSupport elements 
- when the QuestionnaireResponse is valid, the DTR client SHALL indicate that to the user and allow them to mark the QuestionnaireResponse as complete
- If any of the retrieved Questionnaires have an effectivePeriod that ends prior to the current date, then the DTR client SHALL change the status of any retrieved work-in-progress QuestionnaireResponses for the expired Questionnaires to 'stopped' and notify the user that the previously recorded content has expired.
- Client SHALL support saving in-progress questionnaires (we have attestation of saving completed questionnaire)
- When a user fills in a value in a QuestionnaireResponse, the DTR client SHALL populate the information-origin extension
- DTR clients SHALL either provide a PractitionerRole in the SMART App launch of DTR or support transmitting the role by means of the [activeRole](https://build.fhir.org/ig/HL7/davinci-dtr/StructureDefinition-activeRole.html) extension within the Practitioner resource.

rpassas (approved):
I think this attestation-style approach makes sense as a first pass for simulating the server against a Full EHR. I left a comment on clarifying what we do not cover, but I might be confused about what is required of the Full EHR vs just the SMART App.

lib/davinci_dtr_test_kit/docs/dtr_full_ehr_suite_description_v201.md
karlnaden: Most of this will be reviewed / addressed as we dig into the requirements for the g(34/35) criteria and bring requirement coverage analysis to this test kit. I think most of this is fine to address as a part of that work and I'm comfortable with the current limitations section in the full EHR documentation.

> client SHALL visually flag any elements that require adjustment to meet validation rules (or is this SMART APP specific?)
> DTR applications SHALL support rendering according to the mustSupport elements

I think both of these are covered by the limitation section where we say that not all questionnaire capabilities are covered. I would include this as a questionnaire rendering requirement.

> when the QuestionnaireResponse is valid, the DTR client SHALL indicate that to the user and allow them to mark the QuestionnaireResponse as complete

I think this is covered by the attestations

> If any of the retrieved Questionnaires have an effectivePeriod that ends prior to the current date, then the DTR client SHALL change the status of any retrieved work-in-progress QuestionnaireResponses for the expired Questionnaires to 'stopped' and notify the user that the previously recorded content has expired.

Will consider this requirement later.

> Client SHALL support saving in-progress questionnaires (we have attestation of saving completed questionnaire)

Will consider this requirement later.

> When a user fills in a value in a QuestionnaireResponse, the DTR client SHALL populate the information-origin extension

Covered by attestations

> DTR clients SHALL either provide a PractitionerRole in the SMART App launch of DTR or support transmitting the role by means of the [activeRole](https://build.fhir.org/ig/HL7/davinci-dtr/StructureDefinition-activeRole.html) extension within the Practitioner resource.

SMART App-specific, not covered right now.

PR #17, opened 2024-09-19
Release 0.12.0

karlnaden:
# Summary

version and dependency bump


tstrass:
I'm getting a purple error when I think it should be a red error. 
image

Why are we `raise`ing an error here instead of making an assertion? This is definitely not right. That will raise a `RuntimeError`. Importantly, assertions raise Inferno errors.

karlnaden:
Created issue [FI-3153](https://oncprojectracking.healthit.gov/support/browse/FI-3153) to address this later.

tstrass (approved):

PR #19, opened 2024-10-02
FI-3069: Full EHR QuestionnaireResponse verification

tstrass:
# Summary

- Replace attestation tests with tests that use a QuestionnaireResponse input for basic conformance and workflow correctness.
  - Includes adding the necessary methods to `DTRQuestionnaireResponseValidation`. This adds some duplication of logic and error messages between this and the pre-population validation. I tried for a while to refactor things to reduce this duplication, but I was struggling to reach a solution that was adequately clear and simple, so I ultimately decided the duplication was preferable.
- Remove erroneous IG version suffix in origin extension URL
- Rewrite `Fixutres` and `FixtureLoader` modules.
  - This wasn't strictly necessary to fulfill this task, but it came about as part of my attempt to refactor `DTRQuestionnaireResponseValidation`, and I much prefer it this way.
  - `FixtureLoader` has too much logic and information in it. It was meant to be a simple mechanism to minimize file IO during running tests. I moved the test-fixture mapping to `Fixutres` (not the perfect place for it either, but it's fine, see comment).
  - The `Fixtures` interface was unnecessarily complicated so I simplified it to 3 public methods.
  - `Fixtures` now offers class methods rather than instance methods. This is more appropriate given how it works. And I like the added clarity in calling code when you have to reference `Fixtures.` when using it.
- Move Respiratory Assist Device fixtures into a subdirectory like we have with the other fixtures.
- Remove unused group: `client_groups/resp_assist_device/dtr_full_ehr_questionnaire_workflow_group.rb`.
- Improve implementation of `MockPayer#questionnaire_package_response`
  - As discussed offline with @karlnaden.
  - This was also not necessary for this task, but it was something I noticed while refactoring the Fixtures module.

# Testing Guidance

- Run the new tests in the FullEHR suite. Use the QuestionnaireResponses in the Postman collection to test both expected success and expected failures
- Run the SMART App Test Suite to confirm nothing is broken. There shouldn't be any difference in behavior.


karlnaden:
rspec failure due to removed test:

`cannot load such file -- /Users/knaden/Documents/Inferno/source/davinci-dtr-test-kit/lib/davinci_dtr_test_kit/client_groups/dinner_static/dtr_full_ehr_prepopulation_representation_attestation_test
`

lib/davinci_dtr_test_kit/client_groups/dinner_static/dtr_full_ehr_questionnaire_workflow_group.rb
```diff
@@ -6,6 +6,8 @@
 require_relative 'dtr_full_ehr_rendering_enabled_questions_attestation_test'
 require_relative 'dtr_full_ehr_store_attestation_test'
 require_relative 'dtr_full_ehr_prepopulation_representation_attestation_test'
```
karlnaden, line 8: this file is removed - remove this line

tstrass: https://github.com/inferno-framework/davinci-dtr-test-kit/pull/19/commits/4f414d7eeb0f70edf51fe9bb784f9f1f4338d132

lib/davinci_dtr_test_kit/dtr_questionnaire_response_validation.rb
```diff
@@ -1,30 +1,77 @@
...
+    end
+
+    def verify_basic_conformance(questionnaire_response_json)
+      check_is_questionnaire_response(questionnaire_response_json)
+      assert_valid_resource(resource: FHIR.from_contents(questionnaire_response_json),
+                            profile_url: 'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/dtr-questionnaireresponse|2.0.1')
+    end
+
+    # This only checks answers in the questionnaire response, meaning it does not catch missing answers
+    def check_origin_sources(questionnaire_items, response_items, expected_overrides: [])
```
karlnaden, line 30: This method has a lot of similarities with the check_item_prepopulation method, except that the specific values of the answers aren't checked (I think). Is there a reason to keep them separate, or could we have a single set of logic with a flag that turns specific value checking on or off?

tstrass: See PR description:

> This adds some duplication of logic and error messages between this and the pre-population validation. I tried for a while to refactor things to reduce this duplication, but I was struggling to reach a solution that was adequately clear and simple, so I ultimately decided the duplication was preferable.

I really didn't want to add flags to the existing prepopulation validation stuff (`validate_questionnaire_pre_population`) because it's clunky and messy enough as it is. But I'll take another look at just reusing the `check_item_prepopulation` method.

karlnaden (changes requested):
a few small tweaks, but should be close

karlnaden (approved):

PR #22, opened 2024-10-14
FI-3154: Full EHR Adaptive Questionnaire Tests

vanessuniq:
# Summary
This pull request introduces a new set of tests to validate that a DTR Full EHR client can perform a full adaptive questionnaire workflow using the mocked Dinner Order Adaptive Questionnaire. The adaptive workflow involves fetching the adaptive questionnaire package, rendering and pre-populating initial questions, requesting follow-up questions, and completing the adaptive questionnaire response as per the Da Vinci DTR specification.

### Changes Made
* Added **Adaptive Dinner Questionnaire Workflow** for the Full EHR suite, including:
  * Retrieval of the adaptive questionnaire package.
  * Retrieval of next questions.
  * Attestation of proper rendering and population of each next question retrieved.
  * Completion and validation of the `QuestionnaireResponse` for each next-question call.
* Refactored some existing folders and files to allow test reusability and avoid duplication of similar tests. 
* Updated the **Postman Collection** to include new tests and requests needed to validate the adaptive questionnaire workflow.

### What's Left to Do
* [x] Cleanup sample input parameter resources to pass validation (will update the postman collection after)
* [x] Add unit tests
* [x] Add Adaptive Dinner Questionnaire Workflow for the DTR SMART APP suite (Maybe another ticket?)

### Note
The file restructuring affects the following open pull request: [FI-3222: DTR Client Tests: Use Random Number for Attestation Continuation Links](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/20/files).  That pull request should be merged first, and I'll rebase this branch accordingly.

# Testing Guidance
Attached postman collection for testing :
[DTR Full EHR Tests Postman Demo.postman_collection.json](https://github.com/user-attachments/files/17364070/DTR.Full.EHR.Tests.Postman.Demo.postman_collection.json)


karlnaden:
I don't think we can separate the $questionnaire-package and first $next-question calls into separate waits. I expect that typical clients will make almost immediate $next-question calls following the $questionnaire-package call that returns the (empty) adaptive questionnaire. Each other call has a user input step after it, so I think the other divisions are fine.

karlnaden:
I tried running against the payer tests and was not successful due to issues with the payer tests. Created a ticket for that [FI-3303](https://oncprojectracking.healthit.gov/support/browse/FI-3303). We might wait for finalize this IG until that is done as I think it might surface issues with these tests, but more likely we should not wait on it and create follow-up tickets if it does, so let's plan on that.

vanessuniq:
while writing unit test for questionnaire response validation, I determined that this method [`check_answer_presence`](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/22/files#diff-3ddb00d63cd9c03e1b48020ad1e00f0b2bf4fdbc09f687ca7c9a68c11a412782R58) is not doing what it is supposed to do. I might be wrong, or it was not just applicable to my scenario, which is ensuring that all required questions have been answered. I wrote this alternative `check_required_answers_presence` below that method. Please review and let me know if both methods are needed.

tstrass:
> while writing unit test for questionnaire response validation, I determined that this method [`check_answer_presence`](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/22/files#diff-3ddb00d63cd9c03e1b48020ad1e00f0b2bf4fdbc09f687ca7c9a68c11a412782R58) is not doing what it is supposed to do. I might be wrong, or it was not just applicable to my scenario, which is ensuring that all required questions have been answered. I wrote this alternative `check_required_answers_presence` below that method. Please review and let me know if both methods are needed.

Looks to me like these methods are doing the exact same thing. The only difference is that `check_answer_presence` checks for an answer `value` and `check_required_answer_presence` just checks for an `answer`. When I run [this test](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/main/lib/davinci_dtr_test_kit/client_groups/dinner_static/dtr_full_ehr_dinner_static_questionnaire_response_correctness_test.rb#L26), `check_answer_presence` does catch missing answers.

vanessuniq:
> > while writing unit test for questionnaire response validation, I determined that this method [`check_answer_presence`](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/22/files#diff-3ddb00d63cd9c03e1b48020ad1e00f0b2bf4fdbc09f687ca7c9a68c11a412782R58) is not doing what it is supposed to do. I might be wrong, or it was not just applicable to my scenario, which is ensuring that all required questions have been answered. I wrote this alternative `check_required_answers_presence` below that method. Please review and let me know if both methods are needed.
> 
> Looks to me like these methods are doing the exact same thing. The only difference is that `check_answer_presence` checks for an answer `value` and `check_required_answer_presence` just checks for an `answer`. When I run [this test](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/main/lib/davinci_dtr_test_kit/client_groups/dinner_static/dtr_full_ehr_dinner_static_questionnaire_response_correctness_test.rb#L26), `check_answer_presence` does catch missing answers.
> 
> image

Are you using the example sample QuestionnaireResponse from the postman collection?

tstrass:
> Are you using the example sample QuestionnaireResponse from the postman collection?

Yes, but with an answer removed. [qr_sample.json](https://github.com/user-attachments/files/17413942/qr_sample.json)

tstrass:
Ah I see the issue. `check_answer_presence` passes if the entire item is not there. Yeah, you can go ahead and replace what's there with your method.

lib/davinci_dtr_test_kit/client_groups/dinner_static/dtr_full_ehr_dinner_static_questionnaire_response_correctness_test.rb
```diff
@@ -23,7 +23,8 @@ class DTRFullEHRDinnerStaticQuestionnaireResponseCorrectnessTest < Inferno::Test
       # questionnaire = Fixtures.find_questionnaire('DinnerOrderStatic')
       qr = FHIR.from_contents(questionnaire_response)
       check_origin_sources(questionnaire.item, qr.item, expected_overrides: ['PBD.2'])
-      check_answer_presence(qr.item, link_ids: ['PBD.1', 'PBD.2', 'LOC.1'])
+      required_link_ids = extract_required_link_ids(questionnaire.item)
+      check_answer_presence(qr.item, required_link_ids)
```
tstrass, line 27: In this test we are checking the presence of answers based on the workflow rather than what is required in the questionnaire. The workflow technically only asks the user to confirm that the questionnaire is pre-populated (PBD.1, PBD.2, LOC.1), and that rendering is accurate, but it doesn't ask the user to attest that they've completed the questionnaire.

We do say in the description for this group:
> The tester will complete the questionnaire such that a QuestionnaireResponse is stored back into Inferno's EHR endpoint

But that feels a little soft/subtle to me to then fail them if they haven't completed all required fields. Presumably you could save the QuestionnaireResponse to the EHR without completing it.

Although now that I'm thinking about it, I'm wondering if we should actually be checking for presence of all required fields. But @karlnaden approved this as is, so I want to check with Karl on the intent for this test. Should this be all required fields? Should we modify the description or attestations at all?

karlnaden: I agree that we should be checking that all required fields have been populated and attestations should be clear that the quesitonnaire should be completed and stored back to the EHR.

karlnaden:
Did some work towards getting this executing against a local copy of the DTR RI. Currently still working against our private fork, but I hope that we will migrate to the public one and potentially make a PR against it eventually. Being able to interact with the form is crucial to making sure that the workflow and the attestation questions that we ask make sense, so I won't approve this until we've got this working.

Here's where I am:
- I can get the RI to do the first $next-question call
- problems come up on the second $next-question call, specifically around CORS on the request to get Coverage/devreqe0470 - in my current setup, this is in the additional "EHR-available resources (Dinner Adaptive)" input, so there may be an issue with the "access-control-allow-origin" header in some of those cases.

Getting this far required:
1. some fixes to our fork of the DTR smart app - committed to the [smart-launch-adaptive-with-ri](https://gitlab.mitre.org/inferno/dtr/-/tree/smart-launch-adaptive-with-RI?ref_type=heads) branch
2. some fixes to the dtr test kit, specifically around the form of the $next-question responses which are not wrapped in parameters (based on "If there is only one out parameter, which is a Resource with the parameter name "return" then the parameter format is not used, and the response is simply the resource itself." from the [FHIR operations spec](https://build.fhir.org/ig/HL7/sdc/OperationDefinition-Questionnaire-next-question.html) and the [$next-question operation definition](https://hl7.org/fhir/uv/sdc/STU3/OperationDefinition-Questionnaire-next-question.html) from SDC which has a single output parameter named `return`), pushed to this branch.

To run
1. install start up the dtr smart app locally (I start with `NODE_OPTIONS=--openssl-legacy-provider npm start`)
2. start a local copy of the reference server using the current main branch 
3. start the dtr test kit and start a SMART App Launch session
4. Run the adaptive questionnaire workflow group with these inputs
 
- client_id = sample
- launch_uri = http://localhost:3005/launch
- EHR-available resources (Dinner Adaptive) = contents of [inferno-served-instances.json](https://github.com/user-attachments/files/17515138/inferno-served-instances.json)

5. use the link to launch the dtr smart app, which should successfully get to an empty questionnaire with a Next Question button.
6. load up the developer tools in the web browser so you can see requests
7. click the Next Question button, which load the first three questions (name, location, and dinner choice), pre-populating the name question
8. Go back to the Inferno tests and do the attestations to get to the next wait for $next-question dialog
9. go to the dtr smart app, make sure you've filled everything in, and click the Next Question button. For me this displayed "Fetching Resources ..." and hung

Let's get this working the rest of the way!

vanessuniq:
> some fixes to the dtr test kit, specifically around the form of the $next-question responses which are not wrapped in parameters (based on "If there is only one out parameter, which is a Resource with the parameter name "return" then the parameter format is not used, and the response is simply the resource itself." from the [FHIR operations spec](https://build.fhir.org/ig/HL7/sdc/OperationDefinition-Questionnaire-next-question.html) and the [$next-question operation definition](https://hl7.org/fhir/uv/sdc/STU3/OperationDefinition-Questionnaire-next-question.html) from SDC which has a single output parameter named return), pushed to this branch.

I don't think the change made to the test kit is right. "If there is only one out parameter, which is a single Resource with the parameter name "return" then the parameter format is not used, and the response is simply the resource itself." should apply to the server's (inferno) response not the client's input parameter (operation request). We should also note that both the 'IN' and 'OUT' parameters in the [$next-question operation definition](https://hl7.org/fhir/uv/sdc/STU3/OperationDefinition-Questionnaire-next-question.html) have a `1..1` cardinality. 

The initial Inferno's response to next question request is still wrong. currently returning a single parameter named `questionnaire-response`. It should be returning a single parameter named `return`.  But if the cardinality specified in the $next-question operation definition is not relevant then I can simply return the resource.

vanessuniq:
The code change will also make sense if we ignore the specified cardinality for the `in` parameter.

karlnaden:
Not sure I understand your comments here. The code changes I made are in the [logic that builds Inferno's response to $next-question](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/fi-3154-full-ehr-adaptive-questionnaire-tests/lib/davinci_dtr_test_kit/mock_payer.rb#L78). There are two primary changes (I realize I only listed one and messed up a link).

First, make the output the raw QuestionnaireResponse instead of a Parameters. This is allowed because there is only 1 OUT parameter and it is a Resource (QuestionnaireResponse) as described in the [FHIR Operation page](https://hl7.org/fhir/r4/operations.html#executing). The input parameter is not included in the output, so the 1..1 cardinality does not prevent us from returning the raw QuestionnaireResponse.  This is a MAY requirement, so, it is possible that we want to provide a configuration so that testers can choose which way Inferno provides the response. Not sure.

Second, allow the body of the POST to be the raw QuestionnaireResponse instead of a Parameters. This is allowed based on this statement on the [FHIR Operation page](https://hl7.org/fhir/r4/operations.html#request): "If, when invoking the operation, there is exactly one input parameter of type Resource (irrespective of whether other possible parameters are defined), that the operation can also be executed by a POST with that resource as the body of the request (and no parameters on the url)." This could probably have been done in a cleaner way, so refactor if you want, but I think allowing both approaches should be supported.

> The initial Inferno's response to next question request is still wrong. currently returning a single parameter named questionnaire-response. It should be returning a single parameter named return. But if the cardinality specified in the $next-question operation definition is not relevant then I can simply return the resource.

Your experience doesn't match with my code or what I observed when running against the reference implementation locally.  Can you check again and if you're still having problems, let's meet to sync up.

vanessuniq:
>The initial Inferno's response to next question request is still wrong. currently returning a single parameter named questionnaire-response. It should be returning a single parameter named return. But if the cardinality specified in the $next-question operation definition is not relevant then I can simply return the resource.

For the above statement I was referring to my initial implementation.

I understand the changes made based on the statements from the FHIR Operation Page. I just wanted to make sure that the cardinalities specified in the [$next-question operation definition](https://hl7.org/fhir/uv/sdc/STU3/OperationDefinition-Questionnaire-next-question.html) are not relevant and does not add a constrain to what's stated in the FHIR operation page.

Screenshot 2024-10-29 at 5 57 29 PM

Screenshot 2024-10-30 at 8 42 59 AM

karlnaden:
> I just wanted to make sure that the cardinalities specified in the [$next-question operation definition](https://hl7.org/fhir/uv/sdc/STU3/OperationDefinition-Questionnaire-next-question.html) are not relevant and does not add a constrain to what's stated in the FHIR operation page.

It is a somewhat confusing view since they are in the same table, but input and output parameters are considered separately - input parameters appear only in the operation request (body or sometimes in the url) and output parameters appear only in the operation response body. In each case, when there is only 1 resource in the body, it is permissible to send that resource unwrapped without the surrounding parameter boilerplate as described in the links I included above and the cases that we are dealing with here fall into those situations.

vanessuniq:
This section of the IG might explain why the DTR RI  was making the `$next-question` request without `Questionnaire` as the resource type
Screenshot 2024-11-11 at 8 18 20 PM

karlnaden:
Starting with a review of the workflows and user experience when going through these tests. Next, for the Full EHR tests, here's what I see and some suggestions to clean it up

**Full EHR Flow**
- Attest to DTR launch in context of a patient with a full name (to pre-populate)
- Wait test with instructions for making
    - $questionnaire-package call
    - First $next-question call
- Attestation of name pre-population
- Attestation of first name editing (combine with the prior test as in the SMART app tests - can’t edit if it isn’t pre-populated, so they can fail together)
- ?? Instructions to answer the “What do you want for dinner” question
- Wait test with instructions for making
    - Follow-up $next-question call
- ?? Instructions to Answer additional questions if desired?
- Wait test with instructions for making
    - The last $next-question call
- Attest that the QR has been stored for future use
- Verify QR that has been provided in a text box

**Proposal**: add instructions to following wait tests:

- [ ] Follow-up next-question - answer the what do you want for dinner question and then make a next question request
- [ ] Final next-question - answer the remaining questions and then make a final next-question request

**Proposal**: 
- [ ] get rid of the text box for the QuestionnaireResponse and instead verify the QuestionnaireResponse sent in the last $next-question call

**Proposal**: 
- [ ] combine name pre-population and name editing attestations like it done in the SMART App tests for location

vanessuniq:
For the Full EHR, the issue I see with combining name pre-population and name editing attestations, as done in the SMART App tests for location, is that we would no longer have a test solely focused on verifying that the client can pre-populate data.

karlnaden:
image

all three $next-question requests are getting associated with the last $next-question test even though only one is being made during that section. I'm assuming this is due to how we're using tags and that it isn't currently easy to fix because the functionality to do so doesn't exist in core. If that's right, fine to leave.

lib/davinci_dtr_test_kit/client_groups/dinner_adaptive/dtr_adaptive_questionnaire_next_question_request_validation_test.rb
```diff
@@ -0,0 +1,54 @@
...
+               "Input parameter must contain one `parameter:questionnaire-response` slice. Found #{qr_params_count}"
+
+        questionnaire_response =  questionnaire_response_params.first.resource
+        assert_resource_type(:questionnaire_response, resource: questionnaire_response)
+      else
+        assert_valid_resource(
+          resource: input_params,
+          profile_url: 'http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaireresponse-adapt'
+        )
+      end
```
karlnaden, line 51: this ends up doing things that are very similar to DTRAdaptiveQuestionnaireResponseValidationTest. I think they should be combined into a single test, or have the profile verification taken out of this test to avoid the duplication.

karlnaden:
> For the Full EHR, the issue I see with combining name pre-population and name editing attestations, as done in the SMART App tests for location, is that we would no longer have a test solely focused on verifying that the client can pre-populate data.

Fine to keep them separate.

vanessuniq:
> image
> in the SMART app suite, all three $next-question requests are getting associated with the last $next-question test even though only one is being made during that section. I'm assuming this is due to how we're using tags and that it isn't currently easy to fix because the functionality to do so doesn't exist in core. If that's right, fine to leave.
> 
> One additional thing I think it means is that we should have the whole adaptive group be run as group - otherwise, requests from previous runs could get associated incorrectly.

Yup, I updated to make the adaptive workflow run as group.

lib/davinci_dtr_test_kit/client_groups/dinner_adaptive/dtr_adaptive_questionnaire_next_question_request_validation_test.rb
vanessuniq: Updated the description for this test and removed the profile verification for QR.  Let me know if  the update makes sense.

karlnaden:
image

ugh - requests from a prior run still be included. I think this means that the $next-question tag needs to be specific to which wait test is active so that you can distinguish the first, second, and third $next-question requests.

lib/davinci_dtr_test_kit/client_groups/dinner_adaptive/dtr_adaptive_questionnaire_next_question_request_validation_test.rb
```diff
@@ -0,0 +1,64 @@
...
+      assert request.url == next_url, "Request made to wrong URL: #{request.url}. Should instead be to #{next_url}"
+      assert_valid_json(request.request_body)
+      input_params = FHIR.from_contents(request.request_body)
+      assert input_params.present?, 'Request does not contain a recognized FHIR object'
+      assert_valid_resource_type(input_params)
+
+      if input_params.is_a?(FHIR::Parameters)
+        assert_valid_resource(
+          resource: input_params,
+          profile_url: 'http://hl7.org/fhir/uv/sdc/StructureDefinition/parameters-questionnaire-next-question-in'
```
karlnaden, line 52: I don't think we want to do this profile validation. 
1. I suspect no validation will actually be done because the parameter name used in "in" rather than "questionnaire-response" which we and the spec use everywhere else.
2. Even if validation is done, it will just duplicate stuff in the next test if it goes down deeper into the QuestionnaireResponse.

So let's get rid of it.

karlnaden:
I noticed a discrepancy between SMART and Full EHR that got me thinking: The Full EHR next question tests continue immediately. In the SMART tests, we require users to manually continue since they may need to fetch data for pre-population. Even when the returned questionnaire sections don't have prefetch information needed, I think it is right to not make assumptions about what requests the client will make. In the full EHR, inferno is only simulating the payer server, so there’s no data fetch stuff happening, meaning that I think its ok to continue immediately, though it does risk making assumptions (for example, doesn't the RI client make additional $questionnaire-package requests? - what if an EHR wanted to test itself as a complete full EHR with that app). I think I lean towards just making users continue manually for safety, but I'm not fully confident in that.

If we do continue immediately, there’s still an internal discrepancy between the first $questionnaire-package + $next-question waits, which require a user to continue, and the $next-question only waits, which continue automatically. Looks like it may not be easy right now to make them consistent and have all of them continue automatically after the $next-question call, but maybe that can be resolved when Tom converts these to suite endpoints?

Thoughts?

config/DTR Full EHR Tests Postman Demo.postman_collection.json
karlnaden, line 1: These don't fully pass the qustionnaire response validation - please create one that does.

karlnaden (changes requested):
Would like to see a passing example around qustionnaire response verification

karlnaden:
> Either way is fine with me.

I'm feeling consistent and conservative, so let's have them all wait for the user to signal they've made all the requests needed.

lib/davinci_dtr_test_kit/client_groups/dinner_adaptive/dtr_full_ehr_adaptive_questionnaire_request_test.rb
```diff
@@ -0,0 +1,54 @@
...
+              `#{next_url}`.
+
+            - Inferno will respond with the initial set of questions.
+
+          Inferno will wait for both of these requests to be made.
+
+          ### Continuing the Tests
+
+          When both requests have been made, [Click here](#{resume_pass_url}?token=#{access_token}) to continue.
+        )
```
karlnaden, line 50: add request identification details as on the $next-question waits

karlnaden (approved):

PR #24, opened 2024-10-23
FI-3161: ID Token

tstrass:
# Summary

- Add support for `openid` and `fhirUser` scopes.
- Return an ID Token in addition to access token. I followed [this spec](https://openid.net/specs/openid-connect-core-1_0.html#IDToken).
  - ~I chose to always return the ID token, regardless of whether the SUT requested the `openid` and `fhirUser` scopes. This is because it would add complexity to check which scopes they requested, and I didn't see any harm in always returning an ID Token.~

    Update: I added consideration of OpenID scopes to avoid potentially misleading the tester: https://github.com/inferno-framework/davinci-dtr-test-kit/pull/24/commits/221eaf105d347e8f9c948f6c0abe9af39a89eff2
  - This just hard codes in the Practitioner from the DTR bundle on the reference server
  - This doesn't mock any kind of identity provider
- Expose a new JWKS endpoint, in case the client needs to verify the ID token signature to function.

# Testing Guidance

- Exercise the SMART App suite's authorization steps, including fetching the smart configuration, fetching the JWKS, make an authorize request, make a token request
- Run against the SMART App Launch test kit

### DIFF
--- a/lib/davinci_dtr_test_kit/client_groups/resp_assist_device/dtr_resp_questionnaire_package_request_test.rb
+++ b/lib/davinci_dtr_test_kit/client_groups/resp_assist_device/dtr_resp_questionnaire_package_request_test.rb
@@ -38,7 +38,8 @@ module DaVinciDTRTestKit
             Launch. These references help determine the behavior of the app. Referenced instances
             may be provided in the "EHR-available resources" input.
           ),
-          default: JSON.pretty_generate(['Coverage/cov015', 'DeviceRequest/devreqe0470'])
+          default: JSON.pretty_generate([{ reference: 'Coverage/cov015' },
+                                         { reference: 'DeviceRequest/devreqe0470' }])
     input :ehr_bundle,
           optional: true,
           title: 'EHR-available resources (Respiratory Assist Device)',
--- a/lib/davinci_dtr_test_kit/client_groups/resp_assist_device/dtr_resp_questionnaire_package_request_test.rb
+++ b/lib/davinci_dtr_test_kit/client_groups/resp_assist_device/dtr_resp_questionnaire_package_request_test.rb
@@ -38,8 +38,7 @@ module DaVinciDTRTestKit
             Launch. These references help determine the behavior of the app. Referenced instances
             may be provided in the "EHR-available resources" input.
           ),
-          default: JSON.pretty_generate([{ reference: 'Coverage/cov015' },
-                                         { reference: 'DeviceRequest/devreqe0470' }])
+          default: JSON.pretty_generate(['Coverage/cov015', 'DeviceRequest/devreqe0470'])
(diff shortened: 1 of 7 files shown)


tstrass:
Pushed two commits to fix and improve various things with OpenID and verified with the SMART Launch test kit.

Ready for review again @karlnaden

lib/davinci_dtr_test_kit/client_groups/resp_assist_device/dtr_resp_questionnaire_package_request_test.rb
```diff
@@ -38,8 +38,7 @@ class DTRRespQuestionnairePackageRequestTest < Inferno::Test
             Launch. These references help determine the behavior of the app. Referenced instances
             may be provided in the "EHR-available resources" input.
           ),
-          default: JSON.pretty_generate([{ reference: 'Coverage/cov015' },
-                                         { reference: 'DeviceRequest/devreqe0470' }])
+          default: JSON.pretty_generate(['Coverage/cov015', 'DeviceRequest/devreqe0470'])
```
karlnaden, line 41: Looks like this fhirContext format is a difference between [SMART 2.0](https://hl7.org/fhir/smart-app-launch/STU2/scopes-and-launch-context.html#launch-context-arrives-with-your-access_token) and [SMART 2.2](https://hl7.org/fhir/smart-app-launch/STU2.2/scopes-and-launch-context.html#launch-context-arrives-with-your-access_token). I don't think we want to change the default at this time, but we might want to change it based on a suite option around which version of SMART App Launch is in use. But I think we should leave that for a future ticket.

karlnaden:
I need better instructions for how to run this using the postman collection. Specifically, the authorize and token requests I wasn't able to get to work.

karlnaden (changes requested):
see comments

tstrass:
> I need better instructions for how to run this using the postman collection. Specifically, the authorize and token requests I wasn't able to get to work.

In what way were they not working? You should only need to set the base_url variable and then send the requests when the suite is in a wait. If you saw a 404 on the Authorize request(s), that's just because the response to Authorize is a redirect, and Postman attempts to follow the redirect. You can see the 302 if you look in the console.

lib/davinci_dtr_test_kit/client_groups/resp_assist_device/dtr_resp_questionnaire_package_request_test.rb
tstrass: Reverted https://github.com/inferno-framework/davinci-dtr-test-kit/pull/24/commits/fdb786435668713d0718e15e8c1444925e4e36be

karlnaden (approved):

PR #26, opened 2024-11-05
Add USCore 3.1.1 test group, CRD profile tests

elsaperelli:
# Summary
This task adds US Core 3.1.1 tests to the Light DTR EHR Test Suite as well as tests for the [Light DTR EHR REST Capabilities by Resource/Profile](https://hl7.org/fhir/us/davinci-dtr/CapabilityStatement-light-dtr-ehr.html#resource--details). 

Originally, the goal of this task was to generate tests using the US Core generator with a capability statement that combined the US Core 3.1.1 capability statement and the Light DTR EHR Capability statement; however, the majority of the CRD Profiles are read only and the US Core generator read test generator relies on search tests in order to work. Therefore, the majority of the tests are handwritten, with the exception of the CRD Coverage patient and context search tests which were generated and brought over (along with the US Core generator infrastructure).

An additional thing to note is the use of the CDex Task profile in the Task validation test since there is seemingly no CRD Task profile. 

# Testing Guidance
- In order to test the CRD profile tests, I added a patient with CRD profile resources to a local instance of the inferno-reference-server. I have attached this patient JSON file below.
- In order to load the patient JSON into the local copy of the reference server, follow these steps:
    - Put the JSON file in the `/resources` directory
    - `docker compose down`
    - `docker volume rm inferno-reference-server_fhir-pgdata`
    - `docker compose up --build`
- In the Light DTR EHR test kit, use `http://localhost:8080/reference-server/r4` 
- For the profile ids use the following for `pat001`: 
    - `coverage-example1`
    - `communicationrequest-example1`
    - `devicerequest-example1`
    - `encounter-example1`
    - `medicationrequest-example1`
    - `nutritionorder-example1`
    - `servicerequest-example1`
    - `cdex-task-example19`
    - `visionprescription-example`, `visionprescription-example2`

# Questions
- I am a bit confused about the IG's description of the Coverage search by `context` capability. I am not sure how to properly test that since it looks like the example CRD Coverage does not have a `context` field, nor does there seem to be one in the profile Snapshot table. 
- Looks like the Ruby lint errors are coming from the US Core generator files that I had to bring over, what is the best way for me to address these?

[dtr_bundle_patient_pat001.json](https://github.com/user-attachments/files/17680280/dtr_bundle_patient_pat001.json)


lib/davinci_dtr_test_kit/dtr_light_ehr_suite.rb
```diff
@@ -71,5 +104,107 @@ class DTRLightEHRSuite < Inferno::TestSuite
...
+            optional: true
+
+      # All FHIR requests in this suite will use this FHIR client
+      fhir_client do
+        url :url
+        oauth_credentials :credentials
+      end
+
+      group do
+        title 'CRD Coverage'
```
karlnaden, line 130: These should each be [run as group](https://github.com/inferno-framework/us-core-test-kit/blob/main/lib/us_core_test_kit/generated/v5.0.1/condition_encounter_diagnosis_group.rb#L80) since they assume scratch is populated by the first test.

karlnaden:
> I am a bit confused about the IG's description of the Coverage search by context capability. I am not sure how to properly test that since it looks like the example CRD Coverage does not have a context field, nor does there seem to be one in the profile Snapshot table.

Well, this is interesting. I agree there is no context search parameter on the Coverage resource, not in base FHIR R4 or defined in DTR or CRD. Looking at the [current build version of DTR](https://build.fhir.org/ig/HL7/davinci-dtr/CapabilityStatement-light-dtr-ehr-311.html), it looks like the Coverage resource was a mistake and that QuestionnaireResponse was intended. Some of the text in the published version suggests that as well, e.g. "This requires the server to provide access to the specified resources to allow such an app to retrieve and edit QuestionnaireResponses and related resources." and "[context](http://hl7.org/fhir/R4/coverage.html#search) (reference): Allows retrieving QuestionnaireResponse for a context" and the [context search parameter](https://hl7.org/fhir/us/davinci-dtr/STU2/SearchParameter-dtr-context.html) is defined in the published IG on QuestionnaireResponse.

I think this is probably enough for us to switch from Coverage to QuestionnaireResponse and so I think that is what we should do, but will get some leadership opinions before confirming.

lib/davinci_dtr_test_kit/profiles/questionnaire_response/questionnaire_response_validation.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def scratch_resources
+      scratch[:questionnaire_response_resources] ||= {}
+    end
+
+    run do
+      perform_validation_test(scratch_resources[:all] || [],
+                              'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/dtr-questionnaireresponse',
+                              '2.1.0',
```
karlnaden, line 33: DTR published version is 2.0.1 - this causes validator errors for me because it can't find that profile.

lib/davinci_dtr_test_kit/profiles/questionnaire_response/metadata.yml
```diff
@@ -0,0 +1,167 @@
+---
+:name: dtr_questionnaireresponse
+:class_name: DtrQuestionnaireresponseSequence
+:version: v3.1.1
+:reformatted_version: v311
+:resource: QuestionnaireResponse
+:profile_url: http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/dtr-questionnaireresponse
+:profile_name: DTR Questionnaire Response
+:profile_version: 2.1.0
```
karlnaden, line 9: DTR published version is 2.0.1

lib/davinci_dtr_test_kit/profiles/communication_request/communication_request_validation.rb
```diff
@@ -0,0 +1,34 @@
+require_relative '../../validation_test'
+
+module DaVinciDTRTestKit
+  class CommunicationRequestValidationTest < Inferno::Test
+    include DaVinciDTRTestKit::ValidationTest
+
+    title 'CommunicationRequest resources returned during previous tests conform to the CRD CommunicationRequest'
+    description %(
+This test verifies resources returned from the first read conform to
+the [CRD CommunicationRequest](http://hl7.org/fhir/us/davinci-crd/StructureDefinition/profile-communicationrequest).
```
karlnaden, line 10: let's choose a specific version of the CRD IG to point to: https://hl7.org/fhir/us/davinci-crd/STU2/...  Applies to all of these.

lib/davinci_dtr_test_kit/profiles/communication_request/communication_request_validation.rb
```diff
@@ -0,0 +1,34 @@
...
+    id :communication_request_validation
+    input :communication_request_resources
+
+    def resource_type
+      'CommunicationRequest'
+    end
+
+    run do
+      perform_request_validation_test(communication_request_resources, resource_type,
+                                      'http://hl7.org/fhir/us/davinci-crd/StructureDefinition/profile-communicationrequest',
```
karlnaden, line 30: add specific version so it doesn't choose the latest.  Applies to all of these.

lib/davinci_dtr_test_kit/profiles/task/task_validation.rb
```diff
@@ -0,0 +1,34 @@
+require_relative '../../validation_test'
+
+module DaVinciDTRTestKit
+  class TaskValidationTest < Inferno::Test
+    include DaVinciDTRTestKit::ValidationTest
+
+    title 'Task resources returned during previous tests conform to the CDex Task'
+    description %(
+This test verifies resources returned from the first read conform to
+the [CDex Task](http://hl7.org/fhir/us/davinci-cdex/StructureDefinition/cdex-task-attachment-request).
```
karlnaden, line 10: for CDex, the version should be 2.0.0: https://hl7.org/fhir/us/davinci-cdex/STU2/...

lib/davinci_dtr_test_kit/read_test.rb
```diff
@@ -0,0 +1,15 @@
+module DaVinciDTRTestKit
+  module ReadTest
+    def perform_read_test(resource_ids, resource_type)
+      resources = []
+      resource_ids.each do |id|
+        fhir_read resource_type, id
+
+        assert_response_status(200)
```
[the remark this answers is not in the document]
karlnaden: a good example of simplified version of us core read test that steve wrote is [this crd test](https://github.com/inferno-framework/davinci-crd-test-kit/blob/main/lib/davinci_crd_test_kit/client_tests/client_fhir_api_read_test.rb).

karlnaden (changes requested):
a few details to clean-up, but looking good

karlnaden:
to address the ci Ruby / test failure, upgrade to us core test kit version 0.9.0 per steve.

lib/davinci_dtr_test_kit/profiles/questionnaire_response/questionnaire_response_validation.rb
elsaperelli: Ah that makes a lot more sense now!

lib/davinci_dtr_test_kit/profiles/communication_request/communication_request_validation.rb
elsaperelli: What should I do about `QuestionnaireResponse`?

lib/davinci_dtr_test_kit/dtr_light_ehr_suite.rb
karlnaden: these other groups still need `run_as_group` like QuestionnaireResponse

lib/davinci_dtr_test_kit/profiles/nutrition_order/nutrition_order_validation.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    id :nutrition_order_validation
+    input :nutrition_order_resources
+
+    def resource_type
+      'NutritionOrder'
+    end
+
+    run do
+      perform_request_validation_test(nutrition_order_resources, resource_type,
```
karlnaden, line 29: looks like this method adds errors to the message set, but doesn't cause the inferno test to fail, so we need to do that step after it. Will be similar to [this code](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/355816d3c02ed74baa7356e9edce628645740c3d/lib/davinci_dtr_test_kit/payer_server_groups/payer_server_next_request_validation_test.rb#L55), but should be a failure, not a skip.

karlnaden: I'd like to create a separate method for this and then create a unit test for it.  The idea is that you can create an [anonymous class](https://github.com/inferno-framework/us-core-test-kit/blob/main/spec/us_core/search_test_spec.rb#L24) with a run method and any associated details, and then [run it](https://github.com/inferno-framework/us-core-test-kit/blob/main/spec/us_core/search_test_spec.rb#L107). See [this Subscriptions PR](https://github.com/inferno-framework/subscriptions-test-kit/pull/12) for some other infrastructure I'm trying to setup around unit testing and get feedback on that may be helpful. Note that you'll need to stub out calls to the validator [as done here](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/main/spec/davinci_dtr_test_kit/dtr_payer_server_questionnaire_package_group_static_response_validation_spec.rb#L66).

lib/davinci_dtr_test_kit/read_test.rb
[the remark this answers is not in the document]
karlnaden: let's also try to create a unit test for this one following [the idea in this thread](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/26/files#r1838632965). Here's how you [stub out responses](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/26/files#r1838632965) from the system under test.

lib/davinci_dtr_test_kit/profiles/communication_request/communication_request_validation.rb
karlnaden: add specific CRD version here and anywhere else you have a crd URL, should be https://hl7.org/fhir/us/davinci-crd/STU2

karlnaden:
Assuming that we're going to add sample instances to the hosted Inferno Reference Server so that we can demonstrate these tests, I think we should clean up the errors and update the [bundle associated with this pull request](https://github.com/user-attachments/files/17680280/dtr_bundle_patient_pat001.json).  I'm seeing:
- QuestionnaireResponse: several errors around example urls.  Here are SNOMED CT codes I think we could use instead:

SNOMED CT: 248218005 - Awake (slightly different from Awake and at rest, but that's ok for our purposes)
https://browser.ihtsdotools.org/?perspective=full&conceptId1=248218005&edition=MAIN/SNOMEDCT-US/2024-09-01&release=&languages=en

SNOMED CT: 1230326008 - Compressed gas
https://browser.ihtsdotools.org/?perspective=full&conceptId1=1230326008&edition=MAIN/SNOMEDCT-US/2024-09-01&release=&languages=en

SNOMED CT: 261382003 - Mask
https://browser.ihtsdotools.org/?perspective=full&conceptId1=261382003&edition=MAIN/SNOMEDCT-US/2024-09-01&release=&languages=en

- CommunicationRequest: this has something to do with the cross version extension used (and required) by the profile. Somehow, the IG build filters it out. Probably a separate ticket and question for Dylan.

- CDex Task: I think everything is related to example urls. These are systems for identifiers so can be updated to be something inferno-related, like perhaps the base FHIR url of the hosted reference server.

karlnaden (approved):
small fixes to example execution resources loaded into the reference server, some of which may be other tickets to address in the future.

elsaperelli:
[dtr_bundle_patient_pat001.json](https://github.com/user-attachments/files/17832990/dtr_bundle_patient_pat001.json)

Updated patient bundle to load in the reference server! Only the CDex Task has errors now, creating a ticket to address this.

elsaperelli:
Added the necessary resources for testing this to the existing `dtr_bundle_patient_pat015.json` file: https://github.com/inferno-framework/inferno-reference-server/pull/188. Modified the preset to reflect this.

PR #29, opened 2024-11-21
FI-3299: Fix Questionnaire Package verification in Payer suite

tstrass:
The payer suite assumed the response to the questionnaire-package operation was a Parameters instance, but it can also be a Bundle. This change accounts for both possibilities.

There are a lot of changes here, which is essentially because there were many places in the code that relied on the assumption that the response contained a Parameters instance.

# Summary

- **Refactor parts of `CQLTest` module and scratch usage:** `scratch[:adaptive_responses]`, `scratch[:next_responses]`, and `scratch[:output_parameters]` held any of several types of things: array of $q-p request objects, array of $next-question request objects, or array of Parameters instances. The `CQLTest` module had a bunch of code that tried to work on some or all of these cases. This meant that adding yet another case (a $q-p request that returns a bundle) was very difficult. Ultimately I decided the best thing to do was refactor those parts of the module and scratch usage. The module is now ~100 lines shorter. Here is a summary of the changes I made to the scratch objects:
  - `scratch[:adaptive_responses]` -> `scratch[:adaptive_questionnaire_bundles]` (array of bundles)
  - `scratch[:next_responses]` -> `scratch[:next_question_questionnaire_responses]` (array of questionnaire responses)
  - `scratch[:output_parameters]` -> `scratch[:static_questionnaire_bundles]` (array of bundles)
- In `CQLTest`, change library error messages to reference libraries by URL instead of by index (more useful)
- Fix/improve a few of the affected test descriptions

# Testing Guidance
- Run the Payer suite static workflow against the SMART App suite. It should all pass again.
- Run the full Payer suite against the reference implementation. The results should all be the same as before, except 2.11 is now passing. It was failing before, but I believe that was due to a bug that I inadvertently fixed.


lib/davinci_dtr_test_kit/payer_server_groups/adaptive_next_questionnaire_extensions_test.rb
karlnaden, line 1: Noticed that the class for this test is the same as in this [class](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/29/files#diff-ed9483e437cd28198d4e0f930cebb20223395b84fc9552dd92894c0af7993fce). Seems like it should be renamed to be different and match the purpose.

karlnaden (approved):
looks good - a few minor comments to potentially address, but I don't need to look at it again.

lib/davinci_dtr_test_kit/payer_server_groups/adaptive_form_questionnaire_expressions_test.rb
```diff
@@ -12,9 +12,10 @@ class PayerAdaptiveFormExpressionsTest < Inferno::Test
 
     run do
       skip_if retrieval_method == 'Static', 'Performing only static flow tests - only one flow is required.'
-      skip_if scratch[:adaptive_responses].nil?, 'No questionnaire bundle returned.'
-      questionnaire_items_test(scratch[:adaptive_responses], final_cql_test: false)
-      scratch[:adaptive_responses] = nil
+      skip_if scratch[:adaptive_questionnaire_bundles].nil?, 'No questionnaire bundle returned.'
+      questionnaires = extract_questionnaires_from_bundles(scratch[:adaptive_questionnaire_bundles])
+      verify_questionnaire_items(questionnaires)
+      scratch[:adaptive_questionnaire_bundles] = nil
```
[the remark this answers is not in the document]
tstrass: I agree, but this is unchanged from before (the line is changed just because I renamed the scratch object). I believe this is necessary for the tests to flow correctly. I may be missing something, but I think ultimately we could and should be using outputs instead of scratch for this. Potentially something to look into in the future

lib/davinci_dtr_test_kit/payer_server_groups/adaptive_next_questionnaire_extensions_test.rb
tstrass: They are subtly different. One is extensions and the other is expressions

lib/davinci_dtr_test_kit/payer_server_groups/adaptive_form_questionnaire_expressions_test.rb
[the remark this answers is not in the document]
karlnaden: sounds good - we can look at in the future as time allows

lib/davinci_dtr_test_kit/payer_server_groups/adaptive_next_questionnaire_extensions_test.rb
karlnaden: Ah, so there are actually two duplicates:

- PayerStaticFormExtensionsTest in [adaptive_next_questionnaire_extensions_test.rb](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/fi-3299-fix-questionnaire-package-response-verification/lib/davinci_dtr_test_kit/payer_server_groups/adaptive_next_questionnaire_extensions_test.rb) and [static_form_questionnaire_extensions_test.rb](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/fi-3299-fix-questionnaire-package-response-verification/lib/davinci_dtr_test_kit/payer_server_groups/static_form_questionnaire_extensions_test.rb)
- PayerStaticFormExpressionsTest in [adaptive_next_questionnaire_expressions_test.rb](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/fi-3299-fix-questionnaire-package-response-verification/lib/davinci_dtr_test_kit/payer_server_groups/adaptive_next_questionnaire_expressions_test.rb) and [static_form_questionnaire_expressions_test.rb](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/fi-3299-fix-questionnaire-package-response-verification/lib/davinci_dtr_test_kit/payer_server_groups/static_form_questionnaire_expressions_test.rb)

in each case the former should have its class name changed to something about next question. I know these are unrelated to this change, but I think it makes sense to change the class names since we're already touching the files.

PR #30, opened 2024-11-22
Fix typo in bearer token

tstrass:
# Summary

When I fixed the typo in the bearer token for the dinner workflow (https://github.com/inferno-framework/davinci-dtr-test-kit/pull/16), I missed that the same typo was present for the respiratory assist device workflow.

# Testing Guidance

Run the respiratory assist device workflow, make a questionnaire package request using the bearer token displayed in the UI


karlnaden (approved):
works as expected. Doesn't work with the old value.

PR #31, opened 2024-11-23
Add create and update tests for QuestionnaireResponse and Task

elsaperelli:
# Summary
This PR adds write tests (create and update) for DTR QuestionnaireResponse and CDex Task resources in the DTR Light EHR test suite.

# Testing Guidance
- I added example QuestionnaireResponse and Task create and update resources to the `inferno_dtr_server_suite.json` file for ease of testing.
- Using this preset requires a local instance of the inferno reference server. Additionally, you have to allow write interactions in your local inferno reference server instance. You can do this by adding `- READ_ONLY=false` after line 19 in `docker-compose.yml`


lib/davinci_dtr_test_kit/profiles/questionnaire_response/questionnaire_response_create.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module DaVinciDTRTestKit
+  class QuestionnaireResponseCreateTest < Inferno::Test
+    include DaVinciDTRTestKit::CreateTest
+
+    title 'Server is capable of creating a QuestionnaireResponse resource from QuestionnaireResponse create interaction'
+    description 'A sever SHALL support the QuestionnaireResponse create interaction'
+
+    id :questionnaire_response_create
+    input :create_questionnaire_resources
```
karlnaden, line 11: Flesh out this input a bit more - at minimum is should have the textarea type since it is a big block of content. Applies to the other similar create and update inputs.

karlnaden (changes requested):
Looks good - one small adjustment around inputs

karlnaden (approved):

PR #34, opened 2024-12-02
FI-3514: Bug fix for inputs for different workflows

tstrass:
# Summary

Differentiate inputs for different workflows so that you can run static dinner, adaptive dinner, and respiratory assist device workflows in the same test run.

I also threw in some formatting fixes on those inputs. There were a bunch of tick marks intended for code style, but that style doesn't get applied in the frontend.

# Testing Guidance

Click "Run All Tests" at the SMART App suite level. Put something different for each set of inputs, and verify with Postman that the right things get returned at the right time.


karlnaden (approved):
Ran the postman-based demo on all three sets when running together and it worked as expected.

Unit tests would be good to get in place for the Smart App tests at some point, but I don't think now is the right time.

PR #44, opened 2025-01-07
FI-3584: Allow User Provided Questionnaire Package Responses

vanessuniq:
# Summary
This update introduces the ability for users to provide a custom questionnaire package response for the Static Questionnaire workflow test, applicable to both Full EHR and SMART APP Suite. The test logic has been updated to handle the validation of the provided QuestionnaireResponse, and the attestation statements have been adjusted to reflect these changes.

# Test Guidance  

- **Presets for Testing**: Presets include examples like the respiratory device and dinner order questionnaires.  
- **Scenarios to Test**:  
  1. **No Custom Response Provided**: All four validation tests for the custom response should result in "omit."  
  2. **Invalid Custom Response Provided**: If the custom response has no bundle or is of the wrong resource type, the first custom validation should be "fail," and the subsequent three should be "skip."  
  3. **QuestionnaireResponse Validation with Custom Response**: If a custom response is provided but the referenced questionnaire (QuestionnaireResponse.questionnaire) is not in the custom response package, the validation should be skipped.


karlnaden:
Do you have an example questionnaire package response that you've used with this functionality? Walking through an example of how to use it would help with the review.

vanessuniq:
> Do you have an example questionnaire package response that you've used with this functionality? Walking through an example of how to use it would help with the review.

I tested with the [bundle](https://github.com/inferno-framework/davinci-dtr-test-kit/blob/main/lib/davinci_dtr_test_kit/fixtures/respiratory_assist_device/questionnaire_package.json) we use for the  Respiratory Assist Device  workflow and  also this [output parameter](https://hl7.org/fhir/us/davinci-dtr/STU2/Parameters-QuestionnairePackageOutputParamsExample.json.html) from the IG

tstrass:
I think I mostly know what I need to do, but any testing guidance for this?

lib/davinci_dtr_test_kit/client_groups/dinner_static/dtr_full_ehr_questionnaire_workflow_group.rb
```diff
@@ -33,11 +39,46 @@ class DTRFullEHRStaticDinnerQuestionnaireWorkflowGroup < Inferno::TestGroup
...
+      # Test 1: validate the user provided $questionnaire-package response
+      test from: :dtr_custom_questionnaire_package_validation
+      # Test 2: verify the custom response has the necessary libraries for pre-population
+      test from: :dtr_v201_payer_static_form_libraries_test do
+        title 'Custom Questionnaire Package response parameters contain libraries necessary for pre-population'
+        description %(
+          Inferno check that the custom response contains no duplicate library names
+          and that libraries contain cql and elm data.
+        )
+      end
```
tstrass, line 59: image

The results you get in this group are a little confusing if you don't use the custom response. There is a mix of omits and skips, and the messages for the skipped tests ("No questionnaire bundle returned.") don't make sense in this context.

It's nice having the simplicity of just reusing the test from the payer suite (`dtr_v201_payer_static_form_libraries_test`), but I feel like it would be better to just roll a new test, add an `omit_if` like you do in `dtr_custom_questionnaire_package_validation` and then call the same verification method(s) that `dtr_v201_payer_static_form_libraries_test` calls. Omit definitely feels more appropriate here to me than skip.

And if you agree, the same would go for tests 3 and 4 below.

lib/davinci_dtr_test_kit/client_groups/full_ehr/dtr_full_ehr_questionnaire_response_correctness_test.rb
```diff
@@ -1,34 +1,36 @@
...
+                      else
+                        questionnaires = extract_questionnaires_from_bundles(scratch[:static_questionnaire_bundles])
+                        questionnaires.find { |q| qr.questionnaire.end_with?(q.id) }
                       end
-      # questionnaire = Fixtures.find_questionnaire('DinnerOrderStatic')
-      check_origin_sources(questionnaire.item, qr.item, expected_overrides: ['PBD.2'])
+
+      skip_if questionnaire.blank?, "Couldn't find Questionnaire #{qr.questionnaire} to check the QuestionnaireResponse"
+
+      expected_overrides = scratch[:static_questionnaire_bundles].nil? ? [] : ['PBD.2']
```
tstrass, line 30: Shouldn't we no longer be checking the 'PBD.2' override?

lib/davinci_dtr_test_kit/client_groups/full_ehr/dtr_full_ehr_questionnaire_response_correctness_test.rb
```diff
@@ -1,34 +1,36 @@
...
+      skip_if questionnaire_response.blank?, 'Completed QuestionnaireResponse input was blank'
       check_is_questionnaire_response(questionnaire_response)
 
       qr = FHIR.from_contents(questionnaire_response)
-      questionnaire = if config.options[:adaptive]
-                        qr.contained.find do |res|
-                          res.resourceType == 'Questionnaire'
-                        end
-                      else
+      questionnaire = if scratch[:static_questionnaire_bundles].nil?
```
tstrass, line 21: Is the inputted custom questionnaire bundle supposed to be in scratch here? It doesn't seem to be finding it for me

tstrass (changes requested):

lib/davinci_dtr_test_kit/client_groups/dinner_static/dtr_full_ehr_questionnaire_workflow_group.rb
vanessuniq: Makes sense. will make the change

lib/davinci_dtr_test_kit/client_groups/full_ehr/dtr_full_ehr_questionnaire_response_correctness_test.rb
vanessuniq: Actually, I reversed the logic here. What I intended was to check the 'PBD.2' override if a custom response is not provided and skip checking it if a custom response is supplied (`scratch[:static_questionnaire_bundles].nil? ? ['PBD.2'] : []`). We can also choose not to check it at all if that makes more sense.

tstrass: Given this change, I think we should remove the override check

image

lib/davinci_dtr_test_kit/client_groups/full_ehr/dtr_full_ehr_questionnaire_response_correctness_test.rb
vanessuniq: yeah, they don't all run together, so scratch does not persist across groups in this case. Updated the logic, adding the custom response input to this test again

lib/davinci_dtr_test_kit/cql_test.rb
```diff
@@ -320,6 +320,33 @@ def extract_questionnaires_from_bundles(questionnaire_bundles)
...
+        param.resource if param.resource&.resourceType == 'Bundle'
+      end
+    end
+
+    def extract_questionnaire_from_questionnaire_package(questionnaire_pkg_json, questionnaire_id)
+      resource = FHIR.from_contents(questionnaire_pkg_json)
+      questionnaire_bundles = extract_questionnaire_bundles(resource)
+      questionnaires = extract_questionnaires_from_bundles(questionnaire_bundles)
+
+      questionnaires.find { |q| q.id == questionnaire_id }
```
tstrass, line 347: I get this unexpected skip if I use the fixtures for the static dinner questionnaire package and questionnaire response as inputs.
image

I believe it's because the questionnaire reference uses a UUID
https://github.com/inferno-framework/davinci-dtr-test-kit/blob/3e5ef5508de904b888299fa75d39b793594e0518/lib/davinci_dtr_test_kit/fixtures/dinner_static/questionnaire_response_dinner_order_static.json#L155

and the code checks for the questionnaire by `id` rather than `url`.
https://github.com/inferno-framework/davinci-dtr-test-kit/blob/3e5ef5508de904b888299fa75d39b793594e0518/lib/davinci_dtr_test_kit/fixtures/dinner_static/questionnaire_dinner_order_static.json#L9-L10

The QR resource profile requires a `url` so I guess we should be matching on that instead?

tstrass (changes requested):

lib/davinci_dtr_test_kit/cql_test.rb
```diff
@@ -320,6 +320,33 @@ def extract_questionnaires_from_bundles(questionnaire_bundles)
       end&.flatten&.compact
     end
 
+    def extract_questionnaire_bundles(resource)
```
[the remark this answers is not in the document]
vanessuniq: yeah, I did not want to redo the custom response validation over again in the the QuestionnaireResponse group. but should be update

lib/davinci_dtr_test_kit/cql_test.rb
tstrass: I'd have to look at the IG at the resource profiles to figure out what's most appropriate, but either the code or the fixtures or both need to be updated

tstrass: Yeah, so [QuestionnaireResponse.questionnaire](https://hl7.org/fhir/us/davinci-dtr/STU2/StructureDefinition-dtr-questionnaireresponse-definitions.html#QuestionnaireResponse.questionnaire) and [Questionnaire.url](https://hl7.org/fhir/us/davinci-dtr/STU2/StructureDefinition-dtr-std-questionnaire-definitions.html#Questionnaire.url) are both required and both canonical, so I think matching on that would be correct

tstrass (approved):

karlnaden:
Seeing this in practice, I don’t think that changing the dinner static tests to accommodate user-specified $questionnapire-package responses is the right approach. In the SMART tests in particular, we’re able to verify significantly more detail when we know the Questionnaire and the data available to pre-populate it and changing the tests to assume that we don’t know that level of detail is a big step back. I think instead, we should make these completely new tests that test the smaller scope in the basic workflows section and move the dinner static ones elsewhere, like into the Questionnaire Functionality Coverage section.

In addition to that, I think that even though we can’t validate that the client does things like pre-populate correctly, I think we can and should change that all of the “source” values are present: auto, manual, and override.

vanessuniq:
Already merged in #50

PR #45, opened 2025-01-13
FI-3232: Add DTR Specific SMART App launch tests

elsaperelli:
# Summary
This PR adds DTR Light EHR specific SMART Launch requirements tests, following the specifications defined in the Launching a DTR SMART App [here](https://hl7.org/fhir/us/davinci-dtr/STU2/specification.html#launching-dtr).

Specifically, these test additions to the Authorization test group, and more specifically the EHR SMART Launch and Standalone SMART Launch, test that `fhirContext` is included when a SMART App is launched and that that `fhirContext` includes references to one Coverage resource and one QuestionnaireResponse OR Task OR Encounter OR CRD-type Request resource (ServiceRequest, CommunicationRequest, DeviceRequest, or MedicationRequest). 

# Testing Guidance
I have outlined detailed on how to test the new tests without a DTR Light EHR implementation in the `dtr_light_ehr_suite_description_v201.md` [here]([lib/davinci_dtr_test_kit/docs/dtr_light_ehr_suite_description_v201.md](https://github.com/inferno-framework/davinci-dtr-test-kit/compare/fi-3232-smart-launch?expand=1#diff-18dbd48f369a2d3b89d5856a9769c6f7c0e66614bc8bfcac898eaa68f9e160ff)).


lib/davinci_dtr_test_kit/dtr_light_ehr_suite.rb
```diff
@@ -125,16 +136,106 @@ class DTRLightEHRSuite < Inferno::TestSuite
...
-      group from: :service_request_group
-      group from: :task_group
-      group from: :vision_prescription_group
+      group from: :questionnaire_response_group,
+            config: {
+              inputs: {
+                id_token: { name: :ehr_id_token },
+                client_id: { name: :ehr_client_id },
+                requested_scopes: { name: :ehr_requested_scopes },
+                access_token: { name: :ehr_access_token },
```
karlnaden, line 145: This works, but it means that
1. only the token from the ehr launch can get used
2. you had to change all of these tests

Instead, I think we should take the approach that (g)(10) takes to this and rename the output so that both use "access_token" and whichever one got run most recently will be used. See https://github.com/onc-healthit/onc-certification-g10-test-kit/blob/main/lib/onc_certification_g10_test_kit/smart_standalone_patient_app_group.rb#L534 for an example

elsaperelli: Addressed in [Change name of output rather than input](https://github.com/inferno-framework/davinci-dtr-test-kit/pull/45/commits/716da6064d168a9174e48baba2b8f8f5c91f65b4)

karlnaden (approved):
Looks good - can you update the description of the PR to get rid of the questions - otherwise good to go!

PR #60, opened 2025-03-06
FI-3705: Enable Users to Provide Custom Adaptive Questionnaires for Testing

vanessuniq:
# Summary  
This update introduces support for testers to provide their own adaptive questionnaires to run tests in both the **SMART App** and **Full EHR suite**. The custom workflow allows testers to supply an adaptive **Questionnaire package** along with a list (JSON array) of Questionnaires to be included in responses to `$next-question` requests.  

As part of this workflow, the client system must demonstrate the ability to:  
- Retrieve the Questionnaire using the `$questionnaire-package` operation.  
- Request and handle multiple `$next-question` responses.  
- Support the tester in completing the questionnaire through multiple `$next-question` interactions by:  
  - Rendering the Questionnaire.  
  - Pre-populating answers where applicable.  
  - Allowing manual entry of responses, including overriding pre-populated answers.  
- Provide a **completed QuestionnaireResponse** with appropriate indicators for pre-populated and manually entered data (**SMART App only**).  

# Testing Guidance  
**Presets for Testing:**  
- Example presets include **dinner order questionnaires** for demonstration. Use the existing adaptive questionnaire Postman collection for each suite to send requests.  

**Scenarios to Test:**  
  - Tests should **fail** if user-provided inputs are invalid.  
  - If an invalid Questionnaire input is provided, Inferno should return an **`OperationOutcome`** resource in response to the `$next-question` request.  
  - Each provided Questionnaire for a `$next-question` request is validated to ensure:  
    1. **All provided Questionnaires have the same ID** as the contained Questionnaire in the **QuestionnaireResponse** of the `$next-question` request.  
    2. **Each provided Questionnaire includes** all previously presented questions, along with the next question(s).  

**Handling of `$next-question` Requests:**  
  - Each `$next-question` request will receive the **next Questionnaire** from the provided list.  
  - If a `$next-question` request is received when the list is empty, Inferno will mark the **QuestionnaireResponse as completed**.


lib/davinci_dtr_test_kit/client_groups/adaptive_questionnaire/custom/dtr_custom_next_question_response_validation_test.rb
```diff
@@ -0,0 +1,90 @@
...
+        custom_questionnaire = FHIR.from_contents(q.to_json)
+        assert custom_questionnaire, "The custom Questionnaire #{q[:id]} provided is not a valid FHIR resource"
+        assert_resource_type(:questionnaire, resource: custom_questionnaire)
+
+        contained_questionnaire = scratch[:contained_questionnaires][index]
+        assert contained_questionnaire, %(
+          Unable to validate next questionnaire `#{q[:id]}` provided: could not find matching contained
+          questionnaire in the $next-question request.
+        )
+        validate_correctness_of_custom_next_questionnaire(custom_questionnaire, contained_questionnaire)
```
karlnaden, line 80: add error messages and move to the next loop interaction instead of assert

lib/davinci_dtr_test_kit/endpoints/mock_payer/next_question_endpoint.rb
```diff
@@ -22,6 +22,21 @@ def make_response
         response.body = build_questionnaire_next_response.to_json
       end
 
+      def custom_questionnaires
+        return @custom_questionnaires unless @custom_questionnaires.nil?
```
karlnaden, line 26: What's the scope of this caching? I'm worried that with multiple sessions going on at once that you would get interactions that would make this not work.  Or even in a single session, if they run it once without making all of the $next-question requests, won't there be some leftover? It feels like a safer way would be to look at tagged requests to determine how many next-question requests have been made. I think the standard load tagged requests method will scope to just the currently executing wait, so it should work like we need it to.  Do you think I'm wrong to be concerned about this?

spec/davinci_dtr_test_kit/dtr_full_ehr_adaptive_initial_retrieval_group_spec.rb
karlnaden, line 1: I think a spec test for the byo adaptive wait test would be helpful to verify things like
1. expected questionnaire is present in each $next-question response
2. the QuestionnaireResponse is completed when no new questionnaire are in the input.
I'm not sure I saw the right behavior when doing a postman-based demo, so having you verify via a spec test would be very helpful

lib/davinci_dtr_test_kit/client_groups/adaptive_questionnaire/custom/dtr_custom_next_question_response_validation_test.rb
karlnaden: nevermind, I see the rescue that handles it now! that's a nicer pattern than I've been using.

lib/davinci_dtr_test_kit/client_groups/adaptive_questionnaire/dtr_adaptive_next_question_request_test.rb
```diff
@@ -66,6 +66,11 @@ def prompt_cont
     end
 
     run do
+      if respond_to?(:custom_next_question_questionnaire)
```
karlnaden, line 69: is this still needed with the update to separate out the adaptive dinner tests?

vanessuniq: nope updated

lib/davinci_dtr_test_kit/endpoints/mock_payer/next_question_endpoint.rb
vanessuniq: This is within the suite endpoint, and the caching scope is limited to a single request. However, this implementation didn’t return the expected questionnaire for each request—it always returned the first questionnaire in the list. `load_tagged_requests` isn't an option here since it's a runnable instance method. I've updated the logic—let me know your thoughts.

vanessuniq: Thank you for the pointers!! updated.

lib/davinci_dtr_test_kit/client_groups/adaptive_questionnaire/custom/dtr_full_ehr_custom_adaptive_workflow_group.rb
```diff
@@ -0,0 +1,128 @@
...
+          - Provide answers for all required items in their contained Questionnaire.
+        )
+        input :custom_next_question_questionnaires
+      end
+      # Test 5: validate the user provided $questionnaire-package response
+      test from: :dtr_custom_qp_validation
+      # Test 6: verify the custom response has the necessary libraries for pre-population
+      test from: :dtr_custom_questionnaire_libraries
+      # Test 7: validate the user provided $next-question questionnaires
+      test from: :dtr_custom_next_questionnaire_validation
```
karlnaden, line 92: I'm seeing a purple error on this test when no $q-p or $next-question requests are made during the test
image

lib/davinci_dtr_test_kit/client_groups/adaptive_questionnaire/custom/dtr_full_ehr_custom_adaptive_workflow_group.rb
```diff
@@ -0,0 +1,128 @@
...
+           },
+           title: 'Launch DTR (Attestation)'
+      # Test 1: Recieve questionnaire-package and next-question requests
+      test from: :dtr_full_ehr_custom_adative_request
+      # Test 2: validate the $questionnaire-package request body
+      test from: :dtr_qp_request_validation
+      # Test 3: validate the $next-question requests body
+      test from: :dtr_adaptive_next_question_request_validation
+      # Test 4: validate the QuestionnaireResponse in the input parameter
+      test from: :dtr_adaptive_response_validation do
```
karlnaden, line 77: I get a purple error when I make more $next-question requests than in the input:
image

karlnaden (changes requested):
encountered some purple errors to fix

lib/davinci_dtr_test_kit/client_groups/adaptive_questionnaire/custom/dtr_full_ehr_custom_adaptive_workflow_group.rb
vanessuniq: updated, should now skip if no requests are made

lib/davinci_dtr_test_kit/client_groups/adaptive_questionnaire/custom/dtr_full_ehr_custom_adaptive_workflow_group.rb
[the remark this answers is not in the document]
vanessuniq: fixed the format

vanessuniq: fixed

lib/davinci_dtr_test_kit/client_groups/adaptive_questionnaire/dtr_adaptive_response_validation_test.rb
```diff
@@ -0,0 +1,130 @@
...
+        scratch[:contained_questionnaires] << questionnaire
+
+        verify_contained_questionnaire(questionnaire, index, custom_questionnaires)
+        check_missing_origin_sources(qr) if index == requests.length - 1
+
+        expected_overrides = next_request_tag&.include?('custom') ? [] : ['PBD.2']
+        check_origin_sources(questionnaire.item, qr.item, expected_overrides:)
+
+        required_link_ids = extract_required_link_ids(questionnaire.item)
+        check_answer_presence(qr.item, required_link_ids)
```
karlnaden, line 74: Some of these methods add messages rather than asserting them, meaning that the prefixes don't get added because the rescue block is never triggered. Feels like a fiddly and annoying issue to fix now, but I worry if we don't address it we'll forget about it and never come back.  Do you think there's a relatively clear and easy approach that doesn't involve re-writing those methods, or should we create a ticket and defer it until later?

vanessuniq: updated by adding optional `index` param to those methods

karlnaden (approved):

PR #65, opened 2025-03-17
FI-3746: Use AuthInfo

Jammjammjamm:
# Summary
This branch updates the DTR Test Kit to use AuthInfo.

# Testing Guidance
Everything should work as normal.


config/presets/payer_server_run_against_smart_app_tests.json
```diff
@@ -24,10 +24,13 @@
...
     },
     {
-      "name": "credentials",
+      "name": "smart_auth_info",
       "title": "OAuth Credentials",
-      "type": "oauth_credentials",
-      "value": "{\"access_token\":\"eyJhbGciOiJub25lIn0.eyJpbmZlcm5vX2NsaWVudF9pZCI6InNhbXBsZSJ9\"}"
+      "type": "auth_info",
+      "value": {
+        "auth_type": "public",
```
Jammjammjamm, line 31: I'm not sure what auth type this should actually be.

tstrass: This would be `backend_services`, right? The way it's implemented it just needs a bearer token

lib/davinci_dtr_test_kit/dtr_light_ehr_suite.rb
```diff
@@ -131,15 +159,12 @@ class DTRLightEHRSuite < Inferno::TestSuite
                               )
         run_as_group
 
-        input :smart_credentials,
-              title: 'OAuth Credentials',
-              type: :oauth_credentials,
-              optional: true
+        input :smart_auth_info, type: :auth_info
```
vanessuniq, line 162: I think this should still be optional

Jammjammjamm:
The new version of US Core has been released, so this is ready for a final review.

config/presets/payer_server_run_against_smart_app_tests.json
Jammjammjamm: Yeah, I didn't think it would matter from a technical perspective, but it would be nice for the auth type to make sense for the user.

vanessuniq (commented):
looks good, also needs the new core version with the parsing error fix for issue_time

vanessuniq (approved):

PR #67, opened 2025-03-27
FI-3841: Questionnaire Must Support Workflow

vanessuniq:
# Summary
This branch adds Questionnaire must support Tests to SMART APP and Full EHR suite

# Testing Guidance
This is similar to the custom workflow where the user provides the questionnaires. Tested with the dinner order sample data.


lib/davinci_dtr_test_kit/client_groups/must_support/dtr_full_ehr_ms_adaptive_request_test.rb
```diff
@@ -0,0 +1,147 @@
...
+      This test waits for client requests to retrieve and progress through an adaptive questionnaire workflow.
+
+      1. **Questionnaire Package Request**: The client should invoke the `$questionnaire-package` operation
+         to retrieve the adaptive questionnaire package. Inferno will respond with the user-provided
+         empty adaptive questionnaire.
+
+      2. **Next Question Requests**: The client should invoke the `$next-question` operation to request
+         the next set of questions. Inferno will respond sequentially with the next Questionnaire from
+         the user-provided list. If a `$next-question` request is received when the list is empty,
+         Inferno will mark the `QuestionnaireResponse` as completed.
```
karlnaden, line 19: Seems like there can only really be one adaptive questionnaire, since you have to get to the end of the list to have it marked complete... Though I suppose that you don't need to complete the questionnaire to do this demonstration, but its going to make getting the interactions right tricky.  More of an argument for a single test where most things are demonstrated on static questionnaires (there aren't actually any optional must support elements that are unique to adaptive as far as I can tell).

Two options I see here:
1. continue on the current path and update this to be very clear (pithily!) about what's going to happen when multiple adaptive questionnaires are included and $next-question responses for both are in the queue.
2. simplify by having only the static must support test and explain that all questionnaire optional, must support elements can be demonstrated using static questionnaires and leave it at that.

The simplicity of option 2 is seductive. Thoughts?

lib/davinci_dtr_test_kit/client_groups/must_support/dtr_full_ehr_questionnaire_must_support_group.rb
```diff
@@ -0,0 +1,282 @@
...
+      # Test 4: must support test
+      test from: :dtr_questionnaire_must_support,
+           title: '[USER INPUT VALIDATION] All must support elements are provided in the static Questionnaire resources provided', # rubocop:disable Layout/LineLength
+           description: %(
+            The DTR client SHALL be able to handle all `mustSupport` elements defined in the
+            [DTR Standard Questionnaire ](http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/dtr-std-questionnaire)
+            profile. This test will look through the provided Questionnaire resources for the following
+            must support elements:
+
+            - Questionnaire.url
```
karlnaden, line 82: define this list once and pull it from somewhere.

lib/davinci_dtr_test_kit/client_groups/must_support/dtr_questionnaire_must_support_test.rb
```diff
@@ -0,0 +1,42 @@
+require_relative '../../cql_test'
+
+module DaVinciDTRTestKit
+  class DTRQuestionnaireMustSupportTes < Inferno::Test
+    include DaVinciDTRTestKit::CQLTest
+    id :dtr_questionnaire_must_support
+
+    config(options: { debug_must_support_metadata: true })
```
karlnaden, line 8: leftover debug config that should be removed?

karlnaden (changes requested):

karlnaden (approved):

PR #83, opened 2026-04-15
Pre v2.2.0 reorg

karlnaden:
# Summary

Separate tests into folders by actor.

# Testing Guidance

Verified that tests still run, specifically the Full EHR and Payer server tests that are the current focus.


karlnaden:
wrong target

PR #85, opened 2026-05-26
Gem spec updates and Specifying US snomedCT versions in fhir_resource_validator definitions

FlexonyoPizza:
# Summary
- Gem spec updates to new releases of inferno-core (1.3.1), smart-app-launch (1.0.2), tls-test-kit (1.0.2), udap (0.12.2), US-core (1.1.3)
- Bumped jwt to 3.2, no vulnerabilities or rspec failures detected after bumping 
- specified the US snomedCT version in fhir_resource_validator definitions in all 4 suite files


davinci_dtr_test_kit.gemspec
```diff
@@ -8,12 +8,12 @@ Gem::Specification.new do |spec|
   spec.description   = 'Test Kit for the Da Vinci Documentation Templates and Rules (DTR) FHIR Implementation Guide'
   spec.homepage      = 'https://github.com/inferno-framework/davinci-dtr-test-kit'
   spec.license       = 'Apache-2.0'
-  spec.add_dependency 'inferno_core', '~> 1.0', '>= 1.1.2'
-  spec.add_dependency 'jwt', '~> 2.6'
-  spec.add_dependency 'smart_app_launch_test_kit', '~> 1.0'
-  spec.add_dependency 'tls_test_kit', '~> 1.0'
-  spec.add_dependency 'udap_security_test_kit', '~> 0.12.0'
-  spec.add_dependency 'us_core_test_kit', '~> 1.0'
+  spec.add_dependency 'inferno_core', '~> 1.2', '>= 1.2.2'
```
karlnaden: we now have a inferno_core 1.3.1, so might as well update to that.

karlnaden (changes requested):
looking good - those are the right validator changes. Will plan to run some smoke tests myself later today or tomorrow.

lib/davinci_dtr_test_kit/dtr_light_ehr_suite.rb
```diff
@@ -58,6 +58,10 @@ class DTRLightEHRSuite < Inferno::TestSuite
       exclude_message do |message|
         message.message.match?(/\A\S+: \S+: URL value '.*' does not resolve/)
       end
+
```
karlnaden, line 69: Because this test suite is explicitly importing and verifying the US Core 3.1.1 tests, let's add the US Core 3.1.1 message exclusions [as done in CRD](https://github.com/inferno-framework/davinci-crd-test-kit/blob/support-v2.2.0/lib/davinci_crd_test_kit/client/v2.2.1/crd_client_suite.rb#L132C9-L134C12):


```suggestion
exclude_message do |message|
          US_CORE_3_MESSAGE_FILTERS.any? { |match_template| message.message.match?(match_template) }
        end
        
        validation_context do
        snomedCT '731000124108'
      end
```

In addition, you'll need to
- define `US_CORE_3_MESSAGE_FILTERS` as done [here](https://github.com/inferno-framework/davinci-crd-test-kit/blob/support-v2.2.0/lib/davinci_crd_test_kit/client/v2.2.1/crd_client_suite.rb#L116) - adding the single DTR message suppressed.
- add `require 'us_core_test_kit/generated/v3.1.1/us_core_test_suite'` at the top of the suite file.

karlnaden (changes requested):
was able to run the imported US Core tests on the light DTR EHR suite and found some issues to clean up.

karlnaden (approved):

PR #90, opened 2026-07-28
ID-177: Verify all required questions are answered in $next-question requests

abhinandan2012:
## Summary

Partial implementation of requirement [spec-146](https://hl7.org/fhir/us/davinci-dtr/2.2.0/en/specification.html#ci-c-spec-146): the DTR client must not indicate that the user is ready for the next question until the answers in the current QuestionnaireResponse pass validation rules. This adds a check that every required question in the QuestionnaireResponse sent to `$next-question` has been answered.

## Changes

- Factored the QuestionnaireResponse completeness logic out of the v2.2.0 `$next-question` server simulation into a new shared module, `DaVinciDTRTestKit::QuestionnaireResponseCompleteness` (`lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_completeness.rb`). It exposes `contained_questionnaire`, `questionnaire_response_complete?` and `unanswered_required_link_ids`.
- `MockPayer::FullEHRV220NextQuestionEndpoint` now includes the module rather than carrying its own copy of the logic. Behavior is unchanged, including treating a boolean `false` as a valid answer and only marking a QuestionnaireResponse as `completed` when no new questions were disclosed.
- `DTRFullEHRV220NextQuestionRequestValidationTest` now uses the same module to verify each `$next-question` request body. The QuestionnaireResponse is taken from either the `Parameters` wrapper or a bare QuestionnaireResponse body, and the required items in its contained Questionnaire are checked against the answers present. Unanswered required questions are reported per request:

  ```
  (Request 2) All required questions must be answered before requesting the next question.
  No answer found for required item(s): `PBD.1`, `PBD.2`, and `LOC.1`.
  ```

  A QuestionnaireResponse with no contained Questionnaire reports that the check could not be
  performed. The test declares `verifies_requirements 'hl7.fhir.us.davinci-dtr_2.2.0@spec-146'`.

Note that `enableWhen` conditions are not taken into account, so a required question is expected to
be answered even when its enabling condition is not met. This matches the existing behavior of the
server simulation and is documented in the module and in the test description.

## Tests

New spec `spec/davinci_dtr_test_kit/dtr_full_ehr_v220_next_question_request_validation_test_spec.rb`
covers the initial request with no questions disclosed yet, a fully answered form, an unanswered
required question, request numbering across multiple requests, a bare QuestionnaireResponse body, a
required question nested under an answered question, a missing contained Questionnaire, unanswered
optional questions, and a boolean `false` answer.

```
bundle exec rspec                # 318 examples, 0 failures
bundle exec rubocop lib spec     # 229 files inspected, no offenses detected
```

Requirements coverage is unchanged (`bundle exec inferno requirements check_coverage dtr_full_ehr_v220`
reports the coverage CSV is up to date). spec-146 appears in the existing warning list of requirements
not present in the suite's requirement sets, because `requirement_sets` is still commented out in
`dtr_full_ehr_suite.rb` for the v2.2.0 suite.

## Test IDs

- Inferno test: `dtr_full_ehr_v220_nq_request_validation`, shown as **2.2.3.01 Next Question request is valid**
  - Full ID: `dtr_full_ehr_v220-dtr_full_ehr_basic_workflows-dtr_full_ehr_v220_workflow_adaptive-Group03-dtr_full_ehr_v220_nq_request_validation`
- Group to run: `dtr_full_ehr_v220_workflow_adaptive`, shown as **2.2 Adaptive Questionnaire**
- Unit tests:
  - `bundle exec rspec spec/davinci_dtr_test_kit/dtr_full_ehr_v220_next_question_request_validation_test_spec.rb`
  - `bundle exec rspec spec/davinci_dtr_test_kit/full_ehr_v220_next_question_endpoint_spec.rb`

## Where to verify in the UI

Da Vinci DTR Client Test Suite v2.2.0 -> **2 Basic Workflows** -> **2.2 Adaptive Questionnaire** ->
**2.2.3 $next-question** -> **2.2.3.01 Next Question request is valid** -> **MESSAGES** tab. The new
check appears as an error message prefixed with the request number. The **REQUESTS** tab of the same
test shows the `$next-question` bodies the messages refer to.

Note that with hand made request bodies this test also reports profile conformance errors, so the
result is red either way. The item to verify is the presence or absence of the
"All required questions must be answered" message.

## UI testing steps

1. Start a session on the Da Vinci DTR Client Test Suite v2.2.0 with the SMART Backend Services
   client type, and select the "Run Against the Client Simulation Suite" preset so the Client Id is
   set to `demo`.
2. Run **1 Client Registration** and complete the attestation so the Client Id is available to the
   rest of the session.
3. Select **2.2 Adaptive Questionnaire** and click RUN TESTS. The Questionnaires wait dialog appears.
4. Send a `$questionnaire-package` request followed by two `$next-question` requests, where the second
   one repeats the form returned by the first without answering the questions that were disclosed:

   ```sh
   CID=demo
   BASE="http://localhost:4567/custom/dtr_full_ehr_v220/fhir/Questionnaire"
   curl -s -X POST "$BASE/\$questionnaire-package?session_path=$CID" \
     -H 'Content-Type: application/fhir+json' \
     --data-binary @spec/fixtures/questionnaire_package_input_params_conformant.json > /dev/null
   curl -s -X POST "$BASE/\$next-question?session_path=$CID" \
     -H 'Content-Type: application/fhir+json' \
     --data-binary @spec/fixtures/next_question_initial_input_params_conformant.json > nq1.json
   curl -s -X POST "$BASE/\$next-question?session_path=$CID" \
     -H 'Content-Type: application/fhir+json' --data-binary @nq1.json > /dev/null
   ```

   The `session_path` query parameter identifies the session without an access token. It also causes
   the existing request URL check to report a wrong URL, which is expected with this shortcut.
5. Click the continuation link in the wait dialog, then the rendering attestation link.
6. Open **2.2.3.01** and check the MESSAGES tab. Request 2 reports `PBD.1`, `PBD.2` and `LOC.1` as
   unanswered required questions.
7. To confirm there are no false positives, repeat steps 1 to 6 in a new session, but add answers for
   `PBD.1` and `PBD.2` under the `PBD` group and `LOC.1` under the `LOC` group in the third request
   body. The message is no longer reported.

## Running the client and server suites against each other

The paired presets let the Client Simulator Suite drive the Client Test Suite, which exercises the
passing path of this check without a real client, because the simulated client answers each disclosed
question before requesting the next one.

1. Session A: Da Vinci DTR Client Test Suite v2.2.0, SMART Backend Services client type, preset
   "Run Against the Client Simulation Suite". Run **1 Client Registration**, then select
   **2.2 Adaptive Questionnaire** and click RUN TESTS so Inferno waits for requests.
2. Session B: Da Vinci DTR Client Simulator Suite v2.2.0, preset "Run Against the DTR Client Suite".
   The preset points the Payer FHIR Server Base Url at
   `<base_url>/custom/dtr_full_ehr_v220/fhir` and uses client id `demo`. Run **1 Backend Services**
   to obtain a token, then run **2 Payer serves Questionnaires**, which sends the
   `$questionnaire-package` and `$next-question` requests to session A.
3. Return to session A, click through the attestation dialogs, and review **2.2.3.01**. No
   "All required questions must be answered" message is reported for these requests.

## UI proof

Test 2.2.3.01 reporting the unanswered required questions for the second `$next-question` request:

Next Question request is valid reporting unanswered required questions

Location of the test in the Adaptive Questionnaire group:

Location of test 2.2.3.01 within the Adaptive Questionnaire group

The same test for a flow where the disclosed questions were answered. Only profile conformance errors
remain and the completeness message is not present:

Next Question request is valid with all required questions answered

### DIFF
--- a/lib/davinci_dtr_test_kit/full_ehr/v2.2.0/full_ehr_next_question_endpoint.rb
+++ b/lib/davinci_dtr_test_kit/full_ehr/v2.2.0/full_ehr_next_question_endpoint.rb
@@ -2,6 +2,7 @@ require 'udap_security_test_kit'
 require_relative '../endpoints/mock_payer'
 require_relative '../../cross_suite/fhirpath_utils'
 require_relative '../../cross_suite/response_selection_utils'
+require_relative '../../cross_suite/v2.2.0/questionnaire_response_completeness'
 require_relative '../fixture_loader'
 require_relative '../../tags'
 
@@ -11,6 +12,7 @@ module DaVinciDTRTestKit
       include MockPayer
       include DaVinciDTRTestKit::FhirpathUtils
       include DaVinciDTRTestKit::ResponseSelectionUtils
+      include DaVinciDTRTestKit::QuestionnaireResponseCompleteness
 
       def test_run_identifier
         return request.params[:session_path] if request.params[:session_path].present?
@@ -115,22 +117,20 @@ module DaVinciDTRTestKit
       end
 
       def extract_contained_questionnaire
-        contained_questionnaire = request_questionnaire_response.contained&.find do |contained_resource|
-          contained_resource.is_a?(FHIR::Questionnaire)
-        end
+        questionnaire = contained_questionnaire(request_questionnaire_response)
 
-        unless contained_questionnaire.present?
+        unless questionnaire.present?
           return operation_outcome('error', 'invalid',
                                    'Invalid QuestionnaireResponse for $next-question request: ' \
                                    'no contained Questionnaire resource.')
         end
-        unless contained_questionnaire.url.present?
+        unless questionnaire.url.present?
           return operation_outcome('error', 'invalid',
                                    'Invalid QuestionnaireResponse for $next-question request: ' \
                                    'contained Questionnaire must have a url.')
         end
 
-        contained_questionnaire
+        questionnaire
       end
 
       # ***********************************************************************
@@ -290,7 +290,10 @@ module DaVinciDTRTestKit
       def update_response
         @new_questions_added = false
         add_questions_from_questionnaire_template
-        complete_questionnaire_response if !@new_questions_added && questionnaire_response_completed?
+        if !@new_questions_added &&
+           questionnaire_response_complete?(request_questionnaire, request_questionnaire_response)
+          complete_questionnaire_response
+        end
         request_questionnaire_response
       rescue FhirpathServiceError => e
         raise if template_from_fixture?
@@ -336,25 +339,6 @@ module DaVinciDTRTestKit
       # ***********************************************************************
       # Complete QuestionnaireResponse
       # ***********************************************************************
-
-      def questionnaire_response_completed?
-        request_questionnaire.item.all? do |item|
-          item_not_required_or_completed?(item, request_questionnaire_response.item)
-        end
-      end
-
-      def item_not_required_or_completed?(item, response_item_list)
-        required = (item.required == true)
-        response_item = response_item_list.find { |response_item| response_item.linkId == item.linkId }
-        not_required_or_completed = !required || answer_completed_and_valid?(item, response_item)
-        return false unless not_required_or_completed
-
-        item.item.all? { |sub_item| item_not_required_or_completed?(sub_item, response_item&.item || []) }
-      end
-
-      def answer_completed_and_valid?(_item, response_item)
-        response_item&.answer&.any? { |answer| !answer&.value.nil? }
-      end
 
       def complete_questionnaire_response
         request_questionnaire_response.status = 'completed'
--- a/spec/davinci_dtr_test_kit/full_ehr_v220_next_question_endpoint_spec.rb
+++ b/spec/davinci_dtr_test_kit/full_ehr_v220_next_question_endpoint_spec.rb
@@ -968,7 +968,7 @@ RSpec.describe DaVinciDTRTestKit::MockPayer::FullEHRV220NextQuestionEndpoint, :r
       end
 
       it 'sets the QuestionnaireResponse status to completed when the required item answer value is boolean false' do
-        # Regression guard: answer_completed_and_valid? must use !nil? not .present? so that
+        # Regression guard: QuestionnaireResponseCompleteness#answer_present? must use !nil? not .present? so that
         # false (a valid boolean answer) is not mistaken for a missing answer.
         # Q1 is pre-existing in the contained questionnaire so no new questions are added.
         false_answer_body = FHIR::Parameters.new(
--- a/lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_completeness.rb
+++ b/lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_completeness.rb
@@ -1,0 +1,52 @@
+module DaVinciDTRTestKit
+  # Determines whether a QuestionnaireResponse has answers for all of the required questions in the
+  # Questionnaire that it is based on. Used by Inferno's simulated payer server to decide when a
+  # QuestionnaireResponse for an adaptive form is complete, and by the client tests to verify that a
+  # client does not ask for the next question until the current questions have been answered.
+  #
+  # Note that this logic does not take `enableWhen` conditions into account: a required question is
+  # expected to be answered even when its enabling condition is not met.
+  module QuestionnaireResponseCompleteness
+    # The Questionnaire contained within a QuestionnaireResponse, which for an adaptive form holds
+    # the questions disclosed so far.
+    def contained_questionnaire(questionnaire_response)
+      questionnaire_response.contained&.find { |contained_resource| contained_resource.is_a?(FHIR::Questionnaire) }
+    end
+
+    def questionnaire_response_complete?(questionnaire, questionnaire_response)
+      unanswered_required_link_ids(questionnaire, questionnaire_response).empty?
+    end
+
+    # The link ids of the required questions in the questionnaire that do not have an answer in the
+    # questionnaire response. Questions nested within an unanswered required question are omitted.
+    def unanswered_required_link_ids(questionnaire, questionnaire_response)
+      Array(questionnaire.item).flat_map do |item|
+        unanswered_required_link_ids_for_item(item, questionnaire_response.item)
+      end
+    end
+
+    private
+
+    def unanswered_required_link_ids_for_item(item, response_items)
+      response_item = find_response_item(item, response_items)
+      return [item.linkId] if required_and_unanswered?(item, response_item)
+
+      Array(item.item).flat_map do |sub_item|
+        unanswered_required_link_ids_for_item(sub_item, response_item&.item)
+      end
+    end
+
+    def find_response_item(item, response_items)
+      Array(response_items).find { |response_item| response_item.linkId == item.linkId }
+    end
+
+    def required_and_unanswered?(item, response_item)
+      item.required == true && !answer_present?(response_item)
+    end
+
+    def answer_present?(response_item)
+      # an answer value of `false` is a valid answer, so check for nil rather than presence
+      Array(response_item&.answer).any? { |answer| !answer&.value.nil? }
+    end
+  end
+end
--- a/lib/davinci_dtr_test_kit/full_ehr/v2.2.0/request/dtr_next_question_request_validation_test.rb
+++ b/lib/davinci_dtr_test_kit/full_ehr/v2.2.0/request/dtr_next_question_request_validation_test.rb
@@ -1,10 +1,12 @@
 require_relative '../../../urls'
 require_relative '../../../cross_suite/v2.2.0/multi_request_message_helper'
+require_relative '../../../cross_suite/v2.2.0/questionnaire_response_completeness'
 
 module DaVinciDTRTestKit
   class DTRFullEHRV220NextQuestionRequestValidationTest < Inferno::Test
     include URLs
     include MultiRequestMessageHelper
+    include QuestionnaireResponseCompleteness
 
     id :dtr_full_ehr_v220_nq_request_validation
     title 'Next Question request is valid'
@@ -18,13 +20,49 @@ module DaVinciDTRTestKit
       values. CodeableConcept element bindings will fail if none of their codings have a code/system belonging
       to the bound ValueSet. Quantity, Coding, and code element bindings will fail if their code/system are not found in
       the valueset.
+
+      This test also verifies that all of the required questions in the QuestionnaireResponse provided in the
+      request have been answered, because the client is not allowed to indicate that the user is ready for
+      the next question until the answers to the current QuestionnaireResponse pass validation rules. The
+      required questions are the items marked `required` in the Questionnaire contained within the
+      QuestionnaireResponse. Note that `enableWhen` conditions are not considered, so a required question
+      is expected to be answered even if its enabling condition is not met.
     )
+    verifies_requirements 'hl7.fhir.us.davinci-dtr_2.2.0@spec-146'
 
     def target_tags
       tags = [CLIENT_NEXT_TAG]
       tags << config.options[:dtr_workflow_tag] if config.options[:dtr_workflow_tag].present?
 
       tags
+    end
+
+    def questionnaire_response_from_parameters(input_params)
+      resource = input_params.parameter.find { |param| param.name == 'questionnaire-response' }&.resource
+      resource if resource.is_a?(FHIR::QuestionnaireResponse)
+    end
+
+    def check_required_questions_answered(questionnaire_response, request_index)
+      questionnaire = contained_questionnaire(questionnaire_response)
+      if questionnaire.blank?
+        add_request_message(
+          'error',
+          'Unable to verify that all required questions have been answered: the QuestionnaireResponse ' \
+          'does not include a contained Questionnaire.',
+          request_index
+        )
+        return
+      end
+
+      unanswered_link_ids = unanswered_required_link_ids(questionnaire, questionnaire_response)
+      return if unanswered_link_ids.empty?
+
+      add_request_message(
+        'error',
+        'All required questions must be answered before requesting the next question. No answer found ' \
+        "for required item(s): #{unanswered_link_ids.map { |link_id| "`#{link_id}`" }.to_sentence}.",
+        request_index
+      )
     end
 
     run do
@@ -54,11 +92,14 @@ module DaVinciDTRTestKit
           resource_is_valid?(resource: input_params,
                              profile_url: 'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/dtr-questionnaireresponse-adapt|2.2.0',
                              message_prefix: request_prefix(request_index))
+          check_required_questions_answered(input_params, request_index)
 
         elsif input_params.is_a?(FHIR::Parameters)
           resource_is_valid?(resource: input_params,
                              profile_url: 'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/dtr-next-question-input-parameters|2.2.0',
                              message_prefix: request_prefix(request_index))
+          questionnaire_response = questionnaire_response_from_parameters(input_params)
+          check_required_questions_answered(questionnaire_response, request_index) if questionnaire_response.present?
         else
           add_request_message(
             'error',
--- a/spec/davinci_dtr_test_kit/dtr_full_ehr_v220_next_question_request_validation_test_spec.rb
+++ b/spec/davinci_dtr_test_kit/dtr_full_ehr_v220_next_question_request_validation_test_spec.rb
@@ -1,0 +1,186 @@
+require 'davinci_dtr_test_kit/full_ehr/v2.2.0/request/dtr_next_question_request_validation_test'
+
+RSpec.describe DaVinciDTRTestKit::DTRFullEHRV220NextQuestionRequestValidationTest do # rubocop:disable RSpec/SpecFilePathFormat
+  let(:suite_id) { 'dtr_full_ehr_v220' }
+  let(:results_repo) { Inferno::Repositories::Results.new }
+  let(:runnable) { find_test(suite, 'dtr_full_ehr_v220_nq_request_validation') }
+  let(:next_url) { "#{Inferno::Application['base_url']}/custom/#{suite_id}#{DaVinciDTRTestKit::NEXT_PATH}" }
+  let(:request_tags) { [DaVinciDTRTestKit::CLIENT_NEXT_TAG, 'adaptive'] }
+
+  # The initial $next-question request: the contained Questionnaire has no items yet, so there are
+  # no required questions to answer.
+  let(:initial_request_body) do
+    File.read(File.join(__dir__, '..', 'fixtures', 'next_question_initial_input_params_conformant.json'))
+  end
+
+  # A follow-up $next-question request where every required question in the contained Questionnaire
+  # has an answer.
+  let(:all_answered_request_body) do
+    File.read(File.join(__dir__, '..', 'fixtures', 'next_question_input_params_no_origin_extension.json'))
+  end
+
+  # A follow-up $next-question request where required question `3.1` has no answer.
+  let(:missing_answer_request_body) do
+    File.read(File.join(__dir__, '..', 'fixtures', 'next_question_input_params_missing_answer.json'))
+  end
+
+  before do
+    # Profile validation is exercised elsewhere and requires the validator service.
+    allow_any_instance_of(runnable).to receive(:resource_is_valid?).and_return(true)
+  end
+
+  def build_next_requests(*request_bodies)
+    result = repo_create(:result, test_session_id: test_session.id)
+    request_bodies.each do |request_body|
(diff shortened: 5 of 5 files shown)


karlnaden (changes requested):
As you point out, enableWhen behavior can make questions marked as required not actually required, so I think we should check that as a part of this PR. Here's the basic idea, I think

## enableWhen design
- Basic evaluation
    - Find the answer to the specified question
    - Find the compared value (answer)
    - Compare using the operator
    - If the result is true, then consider the question enabled, otherwise don't and it can't be required or otherwise prevent the QuestionnaireResponse from being considered valid and complete.
- Special cases:
    - Children of a question that isn’t enabled are never enabled (no need to check their enableWhen)
    - If multiple enableWhen, assume at least one must evaluate to true (or logic) unless [enableBehavior](https://hl7.org/fhir/us/davinci-dtr/2.2.0/en/StructureDefinition-dtr-base-questionnaire-definitions.html#Questionnaire.item.enableBehavior) (technically required by an invariant) is “all”
    - If a question has multiple answers, evaluate the enableWhen against each individually and consider to be true if at least one evaluates to true
    - Apparently, there can be multiple questions with the same link id (See [question](https://hl7.org/fhir/us/davinci-dtr/2.2.0/en/StructureDefinition-dtr-base-questionnaire-definitions.html#Questionnaire.item.enableWhen.question) element description). Leave this for future work and error if there are multiple questions with the same link id.

karlnaden:
Enable when even more complicated than I thought.  Here's an updated design that takes into account nesting and repetitions that the initial design didn't:

Algorithm for checking required / enabled / present
- Inputs
    - Questionnaire
    - QuestionnaireResponse
- Track
    - Questionnaire node (top or a specific item definition)
    - QuestionnaireResponse node (top, item, or answer, item -> each has a sub “item” element)
    - QuestionnaireResponse ancestors (parent items, each with single answers if inside an answer)
- Logic
    1. Start at top-levels the Questionnaire and QuestionnaireResponse as the current nodes with QuestionnaireResponse ancestors = nil
    2. For each item defined at this node, 
        1. Determine if the linkId is present at this level in the QuestionnaireResponse
            1. Present for group = child item of this QuestionnaireResponse node with the target linkId with a present descendant (a question answer somewhere)
            2. Present for a non-group =child item of this QuestionnaireResponse node with the target linkId and one or more answers
        2. Regardless of whether one found, If there is an enableWhen condition on the Questionnaire, evaluate it (otherwise considered enabled)
            1. If not present in the QuestionnaireResponse, need to find the position where the answer would be in the QuestionnaireResponse based on the question definition order in the Questionnaire
            2. Find the answer for the referenced linkId within the QuestionnaireResponse and using the first item found by searching in this order
                1. First look back through the QuestionnaireResponse ancestors
                2. If no match found, look through all proceeding answers/items (Depth first -> previous sibling (answer) -> its descendants in reverse order, when done with all siblings hop up a level, etc. - with position in current QuestionnaireResponse node shouldn’t need the Questionnaire, but may need to figure out how to keep all answers in the ancestors list while also distinguishing the current answer that we’ve stepped into)
                3. If no match found, look through all following items (Depth first -> next sibling -> its descendants in order, when done with all the siblings hop up a level, etc)
            3. Determine if enabled by evaluating the operator on the whatever was found
                1. Exists -> found or not
                2. Comparator -> if found, one answer from the list must meet the criteria
        3. Evaluate enabled / required
            1. If enabled, required, and not present -> error (required and enabled but not present)
            2. Otherwise if not enabled and present -> error (not enabled, so shouldn’t have been answered)
        4. Recurse: if enabled and present based on the current Questionnaire.item.type
            1. Group means recurse into QuestionnaireResponse’s item.item.
                1. If there are answers -> error (shouldn’t be answers to a group)
                2. Recurse starting at top-level step 2 with the current item as the Questionnaire-level, the response item as the QuestionnaireResponse level, and add the previous QuestionnaireResponse level to the list of ancestors
            2. Non group means recurse into QuestionnaireResponses’ item.answer.item
                1. If there are items -> error (sub-items need to be under the answers) 
                2. Recurse starting at top-level step 2 with the current item as the Questionnaire-level, the response item.answer as the QuestionnaireResponse level, and add the previous QuestionnaireResponse level to the list of ancestors

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
```diff
@@ -0,0 +1,211 @@
...
+    def check_missing_item(item, response_node, definition_order, ancestors:, path:)
+      index = insertion_index(item, response_node, definition_order)
+      return unless item_enabled?(item, QuestionnaireResponsePosition.within(response_node, index), ancestors)
+
+      add_finding(:required_unanswered, item, path) if item.required == true
+      return unless descend_into_absent_group?(item)
+
+      absent_group = QuestionnaireResponseNode.absent_item(item.linkId,
+                                                           parent: response_node, index_in_parent: index)
+      check_group_children(item, absent_group, item.linkId, ancestors:, path:)
```
karlnaden, line 71: I don't think we ever want to descend into a missing group. I [checked with the IG authors]([#Da Vinci DTR > Questionnaire.item.required meaning when nested @ 💬](https://chat.fhir.org/#narrow/channel/197320-Da-Vinci-DTR/topic/Questionnaire.2Eitem.2Erequired.20meaning.20when.20nested/near/615899196)) and a required item under a non-required group or item is not required.  So if a required item/group isn't present, it is enough to report that it is missing and not worry about checking its descendants. If an optional group is missing, then the required sub items don't need to be present, so no further checks are needed.

Note that this means that the dinner questionnaires were incorrect because they needed the required flag on the top-level groups. I've added that and committed it to this branch, so you can use that to check your updated logic.

karlnaden (changes requested):
Looks good - one adjustment needed, which is substantial, so we'll plan to include this in the pre-connectathon deployment later this month instead of next week's deployment.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
abhinandan2012: Fixed. descend_into_absent_group? is gone, and the walk now stops at any item that is missing or holds no answer, so a required item/group that isn't present is reported on itself and its descendants are left alone. Also removed the absent_item/virtual? node support that only existed for that descent.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
```diff
@@ -0,0 +1,159 @@
...
+
+    # This node followed by its descendants, in document order
+    def self_and_descendants
+      [self] + children.flat_map(&:self_and_descendants)
+    end
+
+    # This node followed by its descendants, with the children of each node visited in reverse order.
+    # Used when walking backwards from a position so that nearer nodes come first.
+    def self_and_descendants_reversed
+      [self] + children.reverse.flat_map(&:self_and_descendants_reversed)
```
karlnaden: should these be memoized so that some of the work is shared between nodes? Seems like it could avoid a quadratic blowup.

spec/davinci_dtr_test_kit/full_ehr_v220_next_question_endpoint_spec.rb
```diff
@@ -1141,7 +1141,7 @@ def nq_request_with_preexisting_q1(answered:)
       end
 
       it 'sets the QuestionnaireResponse status to completed when the required item answer value is boolean false' do
-        # Regression guard: answer_completed_and_valid? must use !nil? not .present? so that
+        # Regression guard: QuestionnaireResponseCompleteness#answer_present? must use !nil? not .present? so that
```
karlnaden:
```suggestion
        # Regression guard: QuestionnaireResponseCompleteness#answered? must use !nil? not .present? so that
```

karlnaden (changes requested):
two more small tweaks. We'll need to retarget this onto another branch as this one is now done - I'll create one to target.

karlnaden:
I created branch `v220_client_buildout` - re-open this on top of that branch

spec/davinci_dtr_test_kit/full_ehr_v220_next_question_endpoint_spec.rb
abhinandan2012: Applied.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
abhinandan2012: Memoized both, alongside children.

karlnaden (changes requested):
Looking at spec test coverage, I'd like to see a few more tests in the `dtr_full_ehr_v220_next_question_request_validation_test_spec.rb` spec covering additional cases like group_with_answers and items_outside_answer. Add those and then I think this is good to go.

karlnaden (changes requested):
One more case I think we should handle: there is an [enableWhenExpression extension](https://hl7.org/fhir/uv/sdc/STU4/en/StructureDefinition-sdc-questionnaire-enableWhenExpression.html) that also can be used. We don't have a way in Inferno right now to evaluate these expressions, so we can't check these. I think that's ok, but we want to be clear about it. So I think we should detect when there is an enableWhenExpression extension and:
- presume that the client handled that item correctly - if it is answered, presume enabled; if it wasn't answered, presume disabled (never error in the enabled logic)
- log an "info" message indicating that the expression wasn't evaluated.

abhinandan2012:
## UI testing steps of the `enableWhenExpression` handling

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
```diff
@@ -0,0 +1,236 @@
...
+    # Findings
+    # ***********************************************************************
+
+    def add_finding(type, item, path)
+      location = path.empty? ? '' : " within `#{path.join(' > ')}`"
+      severity = INFORMATIONAL_TYPES.include?(type) ? :info : :error
+      finding = Finding.new(type, severity, item.linkId, path.join(' > '),
+                            "Item `#{item.linkId}`#{location} #{DESCRIPTIONS[type]}.")
+      # A repeating item is evaluated once per occurrence, which would otherwise repeat the same note
+      return if severity == :info && @findings.any? { |existing| existing.message == finding.message }
```
karlnaden: Why only for `:info` findings? If there are multiple answers for a disabled item, won't this add duplicate errors? Are there findings for which the message will be the same but the path will be different? Seems like we might want to be a little more careful here to define what is a duplicate.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
```diff
@@ -0,0 +1,161 @@
...
+    def initialize(parent:, before_index:, after_index:)
+      @parent = parent
+      @before_index = before_index
+      @after_index = after_index
+    end
+
+    # The nodes that come before this position, nearest first: each preceding sibling followed by its
+    # descendants in reverse order, then the siblings preceding the parent, and so on outwards.
+    # Ancestors are not included because they are searched separately.
+    def preceding_nodes
```
karlnaden: memoize for cases with multiple enableWhen conditions

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
```diff
@@ -0,0 +1,161 @@
...
+        end
+        index = node.index_in_parent
+        node = node.parent
+      end
+      nodes
+    end
+
+    # The nodes that come after this position, nearest first: each following sibling followed by its
+    # descendants in document order, then the siblings following the parent, and so on outwards.
+    def following_nodes
```
karlnaden: memoize for cases with multiple enableWhen conditions

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
```diff
@@ -0,0 +1,236 @@
...
+                  ancestors: [], path: [])
+      @findings
+    end
+
+    private
+
+    # Checks each item defined at this level of the questionnaire against the matching items at this
+    # level of the response.
+    def check_level(item_definitions, response_node, ancestors:, path:)
+      definition_order = item_definitions.each_with_index.to_h { |item, index| [item.linkId, index] }
```
karlnaden: This is getting re-computed each time check_level is called and at least `check_question_children` is calling it within a loop where the same item_definition is provided each time.  Hoist it outside of this `check_level` method.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/enable_when_comparison.rb
```diff
@@ -0,0 +1,88 @@
...
+        when *ORDERING_COMPARISON_RESULTS.keys
+          values_ordered?(operator, answer_value, expected_value)
+        else
+          false
+        end
+      end
+
+      private
+
+      def values_equal?(answer_value, expected_value)
```
karlnaden: We should add a special case for Date and Times so that details like timezone don't mess up the comparison

karlnaden (changes requested):
Lots of complexity in this one, found a few more potential issues when looked closer.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
abhinandan2012: Memoized. With four conditions on one item, the list is built once instead of four times.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
abhinandan2012: Memoized, same as above.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
abhinandan2012: Hoisted - built once per definition list and cached by identity, so the repeated `check_level` calls reuse it.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/enable_when_comparison.rb
abhinandan2012: Added. Dates and dateTimes compare as moments (offset-aware), times as seconds since midnight, and a partial date reads as its first moment. Also stopped ordering from falling back to string comparison when only one side is temporal, which used to return 10:00:00 < 2026-01-01.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
abhinandan2012: Fixed properly rather than filtered. The real problem was that two occurrences of a repeating item produced identical messages, since `add_finding` used `path` while the occurrence index lived in `segment`. Findings now carry the occurrence label `(Q1[1], Q1[2])`, and a duplicate is defined once for all severities: same type, same item, same path. Multiple answers on one occurrence still report once. The info note moved to `check_level`, so it's recorded once per item definition.

lib/davinci_dtr_test_kit/full_ehr/v2.2.0/full_ehr_next_question_endpoint.rb
```diff
@@ -317,7 +319,10 @@ def replace_tokens_in_item(item)
       def update_response
         @new_questions_added = false
```
karlnaden: Verifying that the memoized fields are not OperationOutcomes feels like a good defensive practice even if the current code flow verifies that before this is called.

```suggestion
        return request_questionnaire if request_questionnaire.is_a?(FHIR::OperationOutcome)
        return request_questionnaire_response if request_questionnaire_response.is_a?(FHIR::OperationOutcome)
        
        @new_questions_added = false
```

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
karlnaden: One area that it looks like we don't check is that the Questionnaire questions haven't been modified by the client. We could add this in a separate ticket, but it feels very related to this one (prevent clients from removing required questions). Can you add checks that the questions in the embedded Questionnaire match those from the fixture Questionnaire provided for $next-question responses?

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
```diff
@@ -0,0 +1,255 @@
...
+
+    # An item with no matching response item is evaluated from the position it would occupy had it been
+    # answered. Reporting that the item itself is missing is enough: `required` only applies once the
+    # parent is present, so nothing within a missing item needs an answer.
+    def check_missing_item(item, response_node, definition_order, ancestors:, path:)
+      index = insertion_index(item, response_node, definition_order)
+      position = QuestionnaireResponsePosition.within(response_node, index)
+      return unless item_enabled?(item, position, ancestors, present: false)
+
+      add_finding(:required_unanswered, item.linkId, path) if item.required == true
```
karlnaden: move the `if item.required == true` up in the method to avoid work checking if the item is enabled when it isn't needed.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
```diff
@@ -0,0 +1,171 @@
...
+      @children ||= build_children
+    end
+
+    def item_children
+      children.select(&:item?)
+    end
+
+    def answer_children
+      children.select(&:answer?)
+    end
```
karlnaden: These could both be memoized to avoid recomputation

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
```diff
@@ -0,0 +1,171 @@
...
+
+    def item_children_with_link_id(target_link_id)
+      item_children.select { |child| child.link_id == target_link_id }
+    end
```
karlnaden: also a target for memoization due to repeated calls. group by link_id once and access.  For example (test please) - 


```suggestion
    def item_children_by_link_id
      @item_children_by_link_id ||= item_children.group_by(&:link_id)
    end

    def item_children_with_link_id(target_link_id)
      item_children_by_link_id[target_link_id] || []
    end
```

karlnaden (changes requested):
Complex feature, so reviews keep turning up additional improvements.

lib/davinci_dtr_test_kit/full_ehr/v2.2.0/full_ehr_next_question_endpoint.rb
abhinandan2012: Applied, thanks.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
abhinandan2012: moved it up

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
abhinandan2012: Both memoized now.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_node.rb
abhinandan2012: Used your version. It's noticeably faster once there are a lot of items at one level.

lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_response_checker.rb
abhinandan2012: Added. Each request's Questionnaire is now compared against the one we returned in the previous `$next-question` response, so removed, added or altered questions get caught - including flipping `required` off or gating a question with a condition.

One gap: the first request has nothing to compare against, since its Questionnaire comes from `$questionnaire-package`. Easy to add if you want it. Happy to split this out into its own ticket too, since it's new behaviour rather than a fix.

karlnaden: let's add that check - get it complete rather than leave this hanging.

abhinandan2012: Added. The first request is now compared against what `$questionnaire-package` returned, matching on the contained Questionnaire's canonical (`url` or `derivedFrom`), so removed or invented questions get caught there too.

PR #91, opened 2026-07-29
FQM-428: Create 2.2 Payer Suite Skeleton

Jammjammjamm:
# Summary
This branch contains a basic skeleton of what the completed 2.2 payer suite will look like.

# Testing Guidance
A placeholder test to perform validation on the `$questionnaire-package` response was added, which may fail, but otherwise the actual tests in the suite haven't changed, they've just been reorganized, so there shouldn't be any changes to the current client simulation.

# Anticipated Provider-side Test Impact
@karlnaden I'd be interested in any feedback regarding how/whether this organization will impact running against the client suite.

### DIFF
--- a/lib/davinci_dtr_test_kit/server/v2.2.0/dtr_payer_server_suite.rb
+++ b/lib/davinci_dtr_test_kit/server/v2.2.0/dtr_payer_server_suite.rb
@@ -79,14 +79,14 @@ module DaVinciDTRTestKit
           title 'Questionnaire Interactions'
 
           group do
-            title 'Request Validation'
-            simulation_verification
-          end
-
-          group do
             title 'Interactions'
             # TODO: make #questionnaire_interaction identify static/adaptive questionnaires
             test from: :dtr_v220_payer_interaction
+          end
+
+          group do
+            title 'Request Validation'
+            simulation_verification
           end
         end
 
@@ -170,6 +170,8 @@ module DaVinciDTRTestKit
 
       group do
         title 'Log Questionnaire Error Support'
+        # TODO
+        # (optionally ?) perform $q-p operation and validate input
 
         # TODO
         # oper-1 - shall support it
--- a/lib/davinci_dtr_test_kit/server/v2.2.0/dtr_payer_server_suite.rb
+++ b/lib/davinci_dtr_test_kit/server/v2.2.0/dtr_payer_server_suite.rb
@@ -5,7 +5,7 @@ module DaVinciDTRTestKit
   module DTRPayerServerV220
     class DTRPayerServerSuiteV220 < Inferno::TestSuite
       id :dtr_payer_server_v220
-      title 'Da Vinci DTR Client Simulator Suite v2.2.0' # Payer Server v2.2.0 Test Suite' - currently no verification, just simulation  # rubocop:disable Layout/LineLength
+      title 'Da Vinci Payer Server Test Suite v2.2.0'
       description File.read(File.join(__dir__, 'dtr_payer_server_suite_description_v220.md'))
 
       links [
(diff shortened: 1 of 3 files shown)


arscan (approved):

lib/davinci_dtr_test_kit/server/v2.2.0/dtr_payer_server_suite.rb
```diff
@@ -64,7 +75,112 @@ class DTRPayerServerSuiteV220 < Inferno::TestSuite
...
         end
 
-        test from: :dtr_v220_payer_interaction
+        group do
+          title 'Questionnaire Interactions'
+
+          group do
+            title 'Request Validation'
+            simulation_verification
+          end
```
karlnaden: I think it is better to do the request validation after the interaction on the actual request that was sent. Two reasons:
1. You might do some sort of instantiation on the request - this is more likely on responses, but you could still imagine wanting to update the dates within the request.
2. If the validator just started, it can block the interaction until it is warmed up, which is annoying.

lib/davinci_dtr_test_kit/server/v2.2.0/dtr_payer_server_suite.rb
```diff
@@ -64,7 +75,112 @@ class DTRPayerServerSuiteV220 < Inferno::TestSuite
...
+          # oper-15 - [NOT TESTED in 2.0] VS with <40 entries SHALL be expanded
+        end
+      end
+
+      group do
+        title 'Log Questionnaire Error Support'
+
+        # TODO
+        # oper-1 - shall support it
+      end
```
karlnaden, line 175: On the client side, we likely need to test the log error with $questionnaire-package (and maybe $next-question) - e.g., the client sends a $questionnaire-package request, the payer sends something specific back that triggers the error, so I think it may be worth having the option for Inferno to send the full set of interactions before the $log-... operation. Maybe it could be an option, but not a requirement?

karlnaden (approved):
Two small things to think about. I think at a high-level this will work.

lib/davinci_dtr_test_kit/server/v2.2.0/dtr_payer_server_suite.rb
Jammjammjamm: We could also put this in the big group with the other operations, or do you expect it to need to be standalone to use with the client suite?

karlnaden: I expect in the client suite that there will be scenarios where it:
1. waits for a $questionnaire-package request (and maybe $next-question)
2. returns something that should trigger an error
3. continues waiting for a $log-... request that indicates that the client determined there was an error.

For running against the server suite, we'd need a simulation that performs steps 1 and 3 and be able to have it execute them in a sequence. I think most designs for the payer tests will work:
1. If $q-p and $log-... are in separate groups, we'd have the client wait and run the two server suite groups in sequence.
2. If the payer suite's main $q-p group has an option to tell it to send a $log-... request, we'd have the client wait and use that option
3. If the payer suite's $log-... group has the option of doing the $q-p interaction before sending the $log-... request, then we'd have the client wait and run that group.

From the perspective of running the client suite against the payer suite, I like the 3rd option best because it keeps those scenarios separate so that we're not having to re-run the same group again or create a new server session to test the log error scenario.  That said, if that doesn't work with your design of the payer tests, then I'm confident we can work around it.

Jammjammjamm: I think the 3rd option will work fine, so I can go with that.

PR #92, opened 2026-08-03
FQM-432: Add basic response validation

Jammjammjamm:
# Summary
This branch adds basic response validation for the `$questionnaire-package` operation to the server suite.

# Testing Guidance
When run against the client suite, the test will fail because the response does not include a QuestionnaireResponse.

# Anticipated Provider-side Test Impact
None

### DIFF
--- a/lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_operation_validation.rb
+++ b/lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_operation_validation.rb
@@ -1,19 +1,8 @@
 # NOTE: this is just copied from 2.0.1 'cql_test', and hasn't been updated yet
 module DaVinciDTRTestKit
   module QuestionnaireOperationValidation
-    STATIC_QUESTIONNAIRE_PACKAGE_BUNDLE_PROFILE = 'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/DTR-QPackageBundle|2.2.0'.freeze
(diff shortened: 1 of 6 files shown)


lib/davinci_dtr_test_kit/cross_suite/v2.2.0/questionnaire_operation_validation.rb
```diff
@@ -22,334 +11,334 @@ def extension_presence
...
-    def validate_questionnaire_package(resource, form)
-      case resource&.resourceType
-      when 'Bundle'
-        assert_valid_resource(resource:, profile_url: questionnaire_package_profile_urls[:"#{form}_bundle"])
-      when 'Parameters'
-        assert_valid_resource(resource:, profile_url: questionnaire_package_profile_urls[:"#{form}_parameter"])
-      else
-        assert(false, "Unexpected resourceType: #{resource&.resourceType}. Expected Parameters or Bundle")
-      end
+    # def cql_presence
```
Jammjammjamm, line 14: This code is commented out to highlight which code in this module is actually being used in this test. The next branch I'm working on will reincorporate this commented code.

spec/davinci_dtr_test_kit/server/v2.2.0/questionnaire_response_validation_test_spec.rb
kjmahalingam, line 1: Do we want a passing result test in here?

Jammjammjamm: Good catch, thanks.

